PR #28, opened 2020-10-04
feat: Connect Postgres to backend

### LINKED ISSUE #8: Setup Postgres DB

DinuWije:
Closes #8


database.go
```diff
@@ -0,0 +1,161 @@
+package main
```
dhillondeep, line 1: delete this file we don't need it

dhillondeep: delete this file we don't need it

go.mod
```diff
@@ -6,8 +6,14 @@ require (
 	github.com/go-chi/chi v4.1.2+incompatible
 	github.com/go-chi/cors v1.1.1
 	github.com/go-chi/render v1.0.1
+	github.com/gorilla/mux v1.8.0
```
dhillondeep: these dependencies seem unnecessary. A good tip is to do:
```
go mod tidy
```
This will make sure dependencies are up to date and removes unnecessary ones

internal/database/database.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"fmt"
+
+	"github.com/uwblueprint/shoe-project/config"
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+)
+
+func Connect() (*gorm.DB, error) {
+	dsn := fmt.Sprintf("user=%s password=%s dbname=%s port=%s sslmode=disable host=%s", config.GetDatabaseUser(), config.GetDatabasePassword(), config.GetDatabaseName(), config.GetDatabasePort(), config.GetDatabaseHost())
```
dhillondeep: this line seems too long. wrap it some where around 100 length but usually just wrap long lines in general since it improved code readability

restapi/api.go
```diff
@@ -6,20 +6,23 @@ import (
...
-type api struct{}
+type api struct {
+	database *gorm.DB
+}
 
-func Router() (http.Handler, error) {
+func Router(db *gorm.DB) (http.Handler, error) {
 	r := server.CreateRouter()
-	api := api{}
+	api := api{database: db}
```
dhillondeep: usually convention in go to write this line as:
```go
api := api{
    database: db,
}
```
In our case there is only one field so it doesn't matter

cmd/server.go
```diff
@@ -26,8 +27,13 @@ var (
 			// create and mount ui handler
 			server.Router.Mount("/", uiHandler("ui/dist"))
 
-			// create and mount api Router
-			apiRouter, err := restapi.Router()
+			db, err := database.Connect()
+
+			if err != nil {
+				logger.Fatalw("failed to connect to databse", "Err", err)
```
dhillondeep: log messages in code are Capitalized so let's keep the same convention

dhillondeep (changes requested):

dhillondeep:
@DinuWije can you cleanup commits and rebase on main? Right now old commits are showing up

s278gupt:
please update repo's readme with instructions on setting up local db (installing `postgres`, etc...) (unless there's some config that does this on build?) and if you're willing, some info on the packages chosen e.g. `li/pq` and `gorm`

dhillondeep (approved):

Issue #22, opened 2020-10-01
Create Docker compose for the project


dhillondeep:

dhillondeep:
closed by #29

Issue #6, opened 2020-09-20
Makefile to simplify workflow


dhillondeep:
- [ ] Install required dependencies
- [ ] Run yarn watcher and go build together
- [ ] Independent go server run
- [ ] Independent node server run

Issue #5, opened 2020-09-20
Dockerize the application


dhillondeep:
- [ ] Dockerfile
- [ ] Docker Compose

PR #29, opened 2020-10-04
build: Add Docker compose

AbhiPrasad:
See README for full details.

We get hot-reloading for go + JS just by running `make`, installing new dependencies could get a little tricky for golang (might have to rebuild image), but should be no problem for the javascript side.

I think this is a good start, we can fix it as we go.

### DIFF
--- a/docker/Dockerfile.prod
+++ b/docker/Dockerfile.prod
@@ -1,0 +1,51 @@
+############################
+# STEP 1 build executable binary
+############################
+FROM golang:alpine AS builder
+
+RUN apk add --update nodejs yarn util-linux
+
+ENV PORT=8080
+ENV USER=appuser
+ENV UID=10001
+
+# See https://stackoverflow.com/a/55757473/12429735RUN
+RUN adduser \
+    --disabled-password \
+    --gecos "" \
+    --home "/nonexistent" \
+    --shell "/sbin/nologin" \
+    --no-create-home \
+    --uid "${UID}" \
+    "${USER}"
+
+WORKDIR /app
+COPY . .
+
+# fetch dependencies
+RUN go mod download
+RUN go mod verify
+RUN cd ui && yarn && yarn build-prod
+
+# build
+RUN go build -ldflags="-w -s" -o /go/bin/shoe-project
+
+#############################
+## STEP 2 build a small image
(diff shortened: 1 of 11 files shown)


AbhiPrasad:
Also, 

Resolves #5 
Resolves #6 
Resolves #22

aspensantos (approved):

docker/Dockerfile.prod
```diff
@@ -0,0 +1,26 @@
+FROM golang:1.15
```
dhillondeep: For production, let's use `multi-stage` builds. We don't need most of the stuff once we get compiled version of the app and assets.

I would recommend https://github.com/GoogleContainerTools/distroless images and especially `gcr.io/distroless/static` because go is static so let's make the image as small as possible. This image is better than `scratch` because it comes with many security aspects built in

docker/Dockerfile.prod
```diff
@@ -0,0 +1,26 @@
...
+
```
dhillondeep, line 5: For security, let's introduce a user so we don't run our app as privileged root. I refer to `Let’s build a more secure docker image` section of the following article and it has served me well. Maybe have a look!
https://medium.com/@chemidy/create-the-smallest-and-secured-golang-docker-image-based-on-scratch-4752223b7324

docker/Dockerfile.prod
```diff
@@ -0,0 +1,26 @@
...
+WORKDIR /app
+COPY . /app
+
+RUN cd ui && yarn && yarn build-prod
+
+RUN go mod download
+RUN go build
+
+EXPOSE 8900
```
dhillondeep: I am pretty sure on prod we would wanna expose `8080` port. We will change the default port of the app using environment variables when app is deployed

dhillondeep:
@AbhiPrasad @m-penny I fixed the issue with assets not showing up! It was basically go using `/` path as entry point whereas we needed executable path. I also cleaned up both `Dockerfile` and added postgres to `docker-compose`. Can you all squash some of your commits to make it clean and review my changes? We can then get this merged!

AbhiPrasad:
Yup I squashed commits and added some info, lmk and we can merge

PR #31, opened 2020-10-07
Adding standardized styling files 

### LINKED ISSUE #26: Add standardized styling
- Create constants for colours, typography, etc in a separate styles folder

lesliexin:
**Summary**
Closes #26. Adding in standard styles files to organize styling. Controls the base styling across the entire application with just a few files which makes refactoring incredibly simple. In addition, the design system is easily reflected in the source code. 

**What has changed:**
- Added `styles` folder
- Added in files for typography, device sizes (responsiveness), and colours 
- Edited the existing components to adopt the new styling

### DIFF
--- a/ui/src/pages/Home.tsx
+++ b/ui/src/pages/Home.tsx
@@ -1,26 +1,22 @@
 import * as React from 'react';
 import styled from "styled-components";
-
-const Title = styled.h1`
-  font-size: 3em;
-  text-align: center;
-  color: palevioletred;
-`;
+import {color} from "../styles";
+import {TitleText} from "../styles/typography";
 
 const Button = styled.button`
   display: flex;    
-  color: palevioletred;
+  color: ${color.accentColor};
   font-size: 1em;
   margin: 1em;
   padding: 0.25em 1em;
-  border: 2px solid palevioletred;
+  border: 2px solid ${color.accentColor};
   border-radius: 3px;
 `;
 
 export const Home = ()=> {
     return (
         <div>
-          <Title>Home</Title> 
+          <TitleText>Home</TitleText> 
           <span> test text hehe</span> 
           <Button>Test Button</Button>
         </div>
--- a/ui/src/pages/ShoeMap.tsx
+++ b/ui/src/pages/ShoeMap.tsx
@@ -1,10 +1,11 @@
 import * as React from 'react';
+import {TitleText} from "../styles/typography";
 
 export const ShoeMap = ()=> {
     return (
         <div>
-          <h1>ShoeMap</h1> 
-          <span> test text fkjbkjbjdbvdjfv</span> 
+          <TitleText>ShoeMap</TitleText> 
+          <span> test text</span> 
         </div>
     );
   }
--- a/ui/src/styles/typography.js
+++ b/ui/src/styles/typography.js
@@ -1,0 +1,21 @@
+import styled from "styled-components";
+import colors from "./colors";
+import { device } from "./device";
+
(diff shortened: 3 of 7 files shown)


ui/src/styles/typography.js
```diff
@@ -0,0 +1,25 @@
...
+  text-align: center;
+  color: ${colours.accentColour};
+
+  @media ${device.mobileS} {
+    font-size: ${smallText};
+  }
+`;
+
+export const typography = {
+    TitleText: TitleText,
```
AbhiPrasad: Hmm why did we decide to export an object this way?

I'd rather just export each component separately and then import the component from `styles/typography`. This works better with tree shaking btw.

ui/src/styles/colours.js
```diff
@@ -0,0 +1,9 @@
+const white = "#FFFFFF";
+const black = "#000000";
+const pink = '#E9A9A9';
+
+const colours = {
```
AbhiPrasad: +1 on this, good stuff.

ui/src/pages/ShoeMap.tsx
```diff
@@ -1,10 +1,11 @@
 import * as React from 'react';
+import {typography} from "../styles";
 
 export const ShoeMap = ()=> {
     return (
         <div>
-          <h1>ShoeMap</h1> 
-          <span> test text fkjbkjbjdbvdjfv</span> 
+          <typography.TitleText>ShoeMap</typography.TitleText>
```
AbhiPrasad: I'd rather the pattern be import this from a file itself, it helps tree shaking + readability.

Destructuring >> object access IMO.

ui/src/pages/Home.tsx
```diff
@@ -1,26 +1,21 @@
...
-  font-size: 3em;
-  text-align: center;
-  color: palevioletred;
-`;
+import {colour, typography} from "../styles";
 
 const Button = styled.button`
   display: flex;    
-  color: palevioletred;
+  color: ${colour.accentColour};
```
AbhiPrasad: color vs colour lol

I'm of the mind to 🇺🇸  everything, but I'm fine with this way too.

lesliexin: loool I thought long and hard about this and went with 🇨🇦 but i'm fine w/ either

ui/src/styles/typography.js
lesliexin: Only reason why is we can easily change the typography of something within the file. What's currently implemented is a bad example but if we had: 
`export const typography = {
    TitleText: LargeBlackOpenSans,
}`
 and wanted to change the style of the title, we could just: 
`export const typography = {
    TitleText: LargeBlueInter,
}`
versus having to change all occurrences throughout the code. Having the exports this way enables us to componentize. 

Buttttt at the same time I do agree that it has its cons.

AbhiPrasad: Ah gotcha, I can see the reasoning then.

I'm of the mind that this project won't get big enough so we probably don't need to prematurely optimize this stuff, even if that means having to find-and-replace like 5-6 component names, but your call, feel free to keep as is.

AbhiPrasad (approved):
lets ship it

ui/src/styles/typography.js
lesliexin: Hmm I agree, I'll change it!

ui/src/pages/Home.tsx
lesliexin: changed my mind 😅

Issue #21, opened 2020-10-01
Create Postgres schema for Author


dhillondeep:

dhillondeep:
**Author has**

- ID (primary key)
- First Name - STRING
- Last Name - STRING
- Bio - TEXT
- Origin - STRING

Issue #19, opened 2020-10-01
Investigate Multimedia storage solution


dhillondeep:
There are two ways to store multimedia:
- Storing blob data in postgres
- Storing objects in cloud storage

There will be following types of multimedia:
- Images
- Video
- Audio

We need to explore:
- Pros and cons of various approaches
- Which cloud provider to use (most likely same as deployment provider)
- Naming and access management

dhillondeep:
Update:

- We will store images and multimedia data on some data storage host
- There will be a link that must be stored in the database

Update 2
- Verifying cost related topics with The Shoe Project client
- Based on ^^ will consider a cloud provider

PR #41, opened 2020-10-08
Author and story tables

### LINKED ISSUE #10: Create Postgres schema for Story

s278gupt:
Closes: #10 #21  (had to be done together, see `associations`)

Uses author and story GORM models to create (thru migrations) the test db tables whenever the server runs. 

- Associations are: `Story BELONGS TO Author` and `Author HAS MANY Story`

Differences from `Story` schema in #10 :
- Omitted `City` field : already exists in `Author`
- Omitted all multimedia fields : to be updated after #19

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -9,6 +9,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/uwblueprint/shoe-project/config"
 	"github.com/uwblueprint/shoe-project/internal/database"
+	"github.com/uwblueprint/shoe-project/internal/database/migrations"
 	"github.com/uwblueprint/shoe-project/restapi"
 	"github.com/uwblueprint/shoe-project/server"
 	"go.uber.org/zap"
@@ -37,7 +38,18 @@ var (
 
 			db, err := database.Connect()
 			if err != nil {
-				logger.Fatalw("Failed to connect to databse", "Err", err)
+				logger.Fatalw("Failed to connect to database", "Err", err)
+			}
+
+			// initialize tables and seed the database (dev mode only)
+			if config.GetMode() == config.MODE_DEV {
+				if err := migrations.CreateTables(db); err != nil {
+					logger.Fatalw("Database table creation failed", "Err", err)
+				}
+
+				if err := migrations.Seed(db); err != nil {
+					logger.Fatalw("Database seed failed", "Err", err)
+				}
 			}
 
 			apiRouter, err := restapi.Router(db)
--- a/internal/database/migrations/migrations.go
+++ b/internal/database/migrations/migrations.go
@@ -1,0 +1,39 @@
+package migrations
+
+import (
+	"math/rand"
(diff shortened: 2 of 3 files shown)


internal/database/models/models.go
```diff
@@ -0,0 +1,23 @@
...
+	OriginCountry string  `gorm:"not null"`
+	CurrentCity   string  `gorm:"not null"`
+	Stories       []Story `gorm:"foreignKey:AuthorID;constraint:OnUpdate:CASCADE,OnDelete:CASCADE;"`
+}
+
+type Story struct {
+	gorm.Model
+	Title    string
+	Content  string `gorm:"type:text"`
+	AuthorID uint
```
AbhiPrasad, line 21: I see in the gorm docs that they use `int` in some examples, will `uint` be fine too?

internal/database/models/models.go
```diff
@@ -0,0 +1,23 @@
...
+	gorm.Model
+	FirstName     string `gorm:"not null"`
+	LastName      string
+	Bio           string  `gorm:"type:text"`
+	OriginCountry string  `gorm:"not null"`
+	CurrentCity   string  `gorm:"not null"`
+	Stories       []Story `gorm:"foreignKey:AuthorID;constraint:OnUpdate:CASCADE,OnDelete:CASCADE;"`
+}
+
+type Story struct {
```
AbhiPrasad, line 17: Shouldn't some of these fields be `not null`? Like Author or Title?

internal/database/migrations/migrations.go
```diff
@@ -0,0 +1,46 @@
...
+	"math/rand"
+	"strconv"
+	"time"
+
+	"github.com/uwblueprint/shoe-project/internal/database/models"
+	"gorm.io/gorm"
+)
+
+func CreateTables(db *gorm.DB) error {
+	return db.AutoMigrate(&models.Story{})
```
AbhiPrasad: Don't need to do the same for author table?

internal/database/models/models.go
s278gupt: `AuthorID` is a the foreign key, so `uint` is more appropriate

internal/database/migrations/migrations.go
s278gupt: weirdly works without it (briefly looked at `AutoMigrate` implementation, but nothing about foreign keys jumped out), but good catch

cmd/server.go
```diff
@@ -37,7 +38,17 @@ var (
...
 			if err != nil {
-				logger.Fatalw("Failed to connect to databse", "Err", err)
+				logger.Fatalw("Failed to connect to database", "Err", err)
+			}
+
+			// initialize tables and seed the database
+			if err := migrations.CreateTables(db); err != nil {
+				panic(err)
+			}
+			err = migrations.Seed(db)
```
dhillondeep: Seeding should only happen on dev environment

PR #45, opened 2020-10-11
Add Linting -  Goimports

### LINKED ISSUE #30: Add linting to backend
- Choose a linter to use for Go
- Update Makefile to lint and verify
- Add `go mod tidy` to Makefile

Shehryar21:
Closes: #30

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ default: app
 setup:
 	(cd ui; yarn)
 	docker build -t shoe_project_image -f ./docker/Dockerfile.dev .
+	go get golang.org/x/tools/cmd/goimports
 
 app:
 	docker-compose up postgres backend frontend
@@ -30,3 +31,6 @@ clean-docker:
 	docker-compose down
 
 clean: clean-docker clean-frontend clean-backend
+
+fmt:
+	goimports -w .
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -2,6 +2,7 @@ default: docker-build
 
 setup:
 	(cd ui; yarn)
+	go get golang.org/x/tools/cmd/goimports
```
AbhiPrasad, line 6: Can we just run `go install` here as it's in the go mod require?

Shehryar21: Yes, fixed it.

AbhiPrasad:
If you hold off for a bit for https://github.com/uwblueprint/shoe-project/pull/48 to get merged, we can add a CI job for linting.

Makefile
dhillondeep: we actually need that `go get` because a `goimports` is not a package used by our application so it should not be a part of `go.mod`

AbhiPrasad: Why not be part of `go.mod`? It shouldn't add any extra bloat to the go binary right?

dhillondeep: No because `goimports` is a binary that you can install through any method. We are just using go to install it and hence technically this is not a part of our module. Also yeah we don't want unnecessary code as well!

dhillondeep (approved):

PR #55, opened 2020-10-15
add create-stories endpoint

### LINKED ISSUE #13: Create Endpoint: CREATE /stories
This endpoint is a bulk create for stories! Since we are not supporting a dashboard, we will manually send a JSON file with all the data and this endpoint will parse it and store in DB

Shehryar21:
closes #13


dhillondeep (changes requested):
Rebase needed and most of the comments are similar to #56

dhillondeep (approved):

PR #56, opened 2020-10-15
add create-author endpoint

### LINKED ISSUE #57: Create Endpoint: /CREATE authors
Creates authors

Shehryar21:
Closes #57

### DIFF
--- a/restapi/api.go
+++ b/restapi/api.go
@@ -1,9 +1,12 @@
 package restapi
 
 import (
+	"encoding/json"
+	"fmt"
 	"net/http"
 
 	"github.com/go-chi/render"
+	"github.com/uwblueprint/shoe-project/internal/database/models"
 	"github.com/uwblueprint/shoe-project/restapi/rest"
 	"github.com/uwblueprint/shoe-project/server"
 	"go.uber.org/zap"
@@ -27,7 +30,29 @@ func Router(db *gorm.DB) (http.Handler, error) {
 	rest.GetHandler(r, "/stories", api.ReturnAllStories)
 	rest.GetHandler(r, "/story/{storyID}", api.ReturnStoryByID)
 
+	rest.PostHandler(r, "/authors", api.authorCreate)
 	return r, nil
+}
+
+func (api api) authorCreate(w http.ResponseWriter, r *http.Request) render.Renderer {
+	// Declare a new Story struct.
+	var s []models.Author
+
+	// Try to decode the request body into the struct. If there is an error,
+	// respond to the client with the error message and a 400 status code.
+	err := json.NewDecoder(r.Body).Decode(&s)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return rest.MsgStatusOK("Error")
+	}
+
+	if err := api.database.Create(&s).Error; err != nil {
+		return rest.ErrInternal(zap.S(), err)
+	}
+
+	// Do something with the Author struct...
+	fmt.Fprintf(w, "Author: %+v", s)
+	return rest.MsgStatusOK("Hello World")
 }
 
 func (api api) health(w http.ResponseWriter, r *http.Request) render.Renderer {
--- a/restapi/api.go
+++ b/restapi/api.go
@@ -1,12 +1,9 @@
 package restapi
 
 import (
-	"encoding/json"
-	"fmt"
 	"net/http"
 
 	"github.com/go-chi/render"
-	"github.com/uwblueprint/shoe-project/internal/database/models"
 	"github.com/uwblueprint/shoe-project/restapi/rest"
 	"github.com/uwblueprint/shoe-project/server"
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


restapi/api.go
```diff
@@ -21,10 +25,31 @@ func Router(db *gorm.DB) (http.Handler, error) {
...
 
 	rest.GetHandler(r, "/health", api.health)
-
+	rest.PostHandler(r, "/authors", api.authorCreate)
 	return r, nil
 }
 
+func (api api) authorCreate(w http.ResponseWriter, r *http.Request) render.Renderer {
+	// Declare a new Story struct.
+	var s []models.Author
```
dhillondeep: try using descriptive names like `author` lol

restapi/api.go
```diff
@@ -21,10 +25,31 @@ func Router(db *gorm.DB) (http.Handler, error) {
...
+
+	// Try to decode the request body into the struct. If there is an error,
+	// respond to the client with the error message and a 400 status code.
+	err := json.NewDecoder(r.Body).Decode(&s)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
dhillondeep: don't do this because we have our wrapper around this functionality.

Instead do:
```go
rest.ErrInvalidRequest(zap.S(), "Invalid payload", err)
```

restapi/api.go
```diff
@@ -21,10 +25,31 @@ func Router(db *gorm.DB) (http.Handler, error) {
...
+		return rest.MsgStatusOK("Error")
```
dhillondeep: ^^similarly remove this

restapi/api.go
```diff
@@ -21,10 +25,31 @@ func Router(db *gorm.DB) (http.Handler, error) {
...
+	}
+
+	if err := api.database.Create(&s).Error; err != nil {
+		return rest.ErrInternal(zap.S(), err)
+	}
+
+	// Do something with the Author struct...
```
dhillondeep: don't have this debug code here!

Instead return a success message:
```go
return rest.MsgStatusOK("Authors added successfully")
```

restapi/api.go
```diff
@@ -21,10 +25,31 @@ func Router(db *gorm.DB) (http.Handler, error) {
...
-
+	rest.PostHandler(r, "/authors", api.authorCreate)
 	return r, nil
 }
 
+func (api api) authorCreate(w http.ResponseWriter, r *http.Request) render.Renderer {
+	// Declare a new Story struct.
+	var s []models.Author
+
+	// Try to decode the request body into the struct. If there is an error,
```
dhillondeep: no need for this comment

restapi/api.go
```diff
@@ -21,10 +25,31 @@ func Router(db *gorm.DB) (http.Handler, error) {
...
+	// respond to the client with the error message and a 400 status code.
+	err := json.NewDecoder(r.Body).Decode(&s)
+	if err != nil {
```
dhillondeep: since we don't need access to any value and we are only checking error, combine lines:

```go
if err := json.NewDecoder(r.Body).Decode(&s); err != nil {
    ...
}; 
```

restapi/api.go
```diff
@@ -21,10 +25,31 @@ func Router(db *gorm.DB) (http.Handler, error) {
 	}
 
 	rest.GetHandler(r, "/health", api.health)
-
+	rest.PostHandler(r, "/authors", api.authorCreate)
 	return r, nil
 }
 
+func (api api) authorCreate(w http.ResponseWriter, r *http.Request) render.Renderer {
```
dhillondeep: naming nitpick but make we should name it `CreateAuthor` this way what we are doing comes first

dhillondeep: If you notice code that is merged in master, I told megan to move her code into `stories.go` file. Can you do the same and move endpoints code into `authors.go`. This way `api.go` won't` be huge

restapi/api.go
```diff
@@ -21,10 +25,31 @@ func Router(db *gorm.DB) (http.Handler, error) {
...
+	// Try to decode the request body into the struct. If there is an error,
+	// respond to the client with the error message and a 400 status code.
+	err := json.NewDecoder(r.Body).Decode(&s)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return rest.MsgStatusOK("Error")
+	}
+
+	if err := api.database.Create(&s).Error; err != nil {
+		return rest.ErrInternal(zap.S(), err)
```
dhillondeep: after rebasing on master, `api` has a field called `logger` so now you can just do `api.logger`

restapi/api.go
```diff
@@ -21,10 +25,31 @@ func Router(db *gorm.DB) (http.Handler, error) {
...
+
+	// Try to decode the request body into the struct. If there is an error,
+	// respond to the client with the error message and a 400 status code.
+	err := json.NewDecoder(r.Body).Decode(&s)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return rest.MsgStatusOK("Error")
+	}
+
+	if err := api.database.Create(&s).Error; err != nil {
```
dhillondeep: what we get is an array of authors so I am pretty sure you will need to look the authors array and do `api.database.Create` on each author. Verify that for me because I could be wrong

dhillondeep (changes requested):

restapi/api.go
Shehryar21: according to the docs, it allows batch insert: https://gorm.io/docs/create.html. I also tested it out by calling the endpoint

dhillondeep: okay sweet if it works less work for us. I will test on my side as well

dhillondeep (approved):

PR #59, opened 2020-10-15
Make migrations and seed into a seperate command

dhillondeep:
- Separate out DB migration and seeding into a different command
- This is because we don't want to seed every time server restarts (which is every time it reloads)
- Made seed data random using random words
- Docker changes to run migration

How to run:
```
./shoe-project seed
```

cc @s278gupt


cmd/seed.go
```diff
@@ -0,0 +1,39 @@
...
+
+			db, err := database.Connect()
+			if err != nil {
+				logger.Fatalw("Failed to connect to database", "Err", err)
+				return
+			}
+
+			if err := migrations.CreateTables(db); err != nil {
+				logger.Fatalw("Database table creation failed", "Err", err)
+			}
```
AbhiPrasad, line 30: Shouldn't we early return here as well?

cmd/seed.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/uwblueprint/shoe-project/internal/database"
+	"github.com/uwblueprint/shoe-project/internal/database/migrations"
+	"go.uber.org/zap"
+)
+
+var (
+	seedCmd = &cobra.Command{
```
AbhiPrasad, line 12: Do we not want to restrict this to running in DEV mode only? (Like we did before by checking config?)

docker-compose.yml
```diff
@@ -1,10 +1,12 @@
 version: '3.7'
+
 services:
   server:
-    container_name: shoe_project_server
     build:
       context: .
       dockerfile: $PWD/docker/Dockerfile.dev
+    container_name: shoe_project_server
+    image: shoe_project_image
```
AbhiPrasad, line 9: can we make sure that `server` depends on `postgres` as well here? Might as well include in this PR tbh.

internal/database/migrations/migrations.go
```diff
@@ -14,23 +11,19 @@ func CreateTables(db *gorm.DB) error {
...
-	rand.Seed(time.Now().Unix())
-	randNum := rand.Int()
+	babbler := babble.NewBabbler()
+	babbler.Count = 1
 
 	seed := models.Author{
-		FirstName:     "Grace " + strconv.Itoa(randNum),
-		LastName:      "Hopper",
-		Bio:           "pioneer",
+		FirstName:     babbler.Babble(),
```
AbhiPrasad: Can we just use https://github.com/dmgk/faker or similar? We can also easily reuse for testing then.

cmd/seed.go
dhillondeep: good catch!

cmd/seed.go
dhillondeep: yeah more I think about it, we should! We can just show an error message if done without `dev` mode

docker-compose.yml
dhillondeep: sure can do

internal/database/migrations/migrations.go
dhillondeep: you are right faker is a pretty good package. I will try integrating it

cmd/seed.go
dhillondeep: `logger.Fatalw` should exit so we don't need `return`

AbhiPrasad (approved):

PR #60, opened 2020-10-15
Make Docker workflow configurable

dhillondeep:
- Add ability to run individual services
- Add documentation regarding workflow

cc @uwblueprint/the-shoe-project


dhillondeep:
Should be merged after #59

docker-compose.yml
```diff
@@ -29,12 +64,16 @@ services:
     container_name: shoe_project_pgadmin
     image: dpage/pgadmin4
     environment:
-      PGADMIN_DEFAULT_EMAIL: user@blueprint.com
-      PGADMIN_DEFAULT_PASSWORD: admin
+      PGADMIN_DEFAULT_EMAIL: 'user@blueprint.com'
```
AbhiPrasad: `uwblueprint.org` is more correct here for us, but I guess it doesn't really matter.

docker-compose.yml
```diff
@@ -1,15 +1,50 @@
...
-    environment:
-      DATABASE_HOST: postgres
+      - '1234:1234'
+    working_dir: '/app/ui'
+    command:
+      - /bin/sh
+      - -c
+      - |
+          yarn watch &
+          yarn start
```
AbhiPrasad, line 46: Why run both `yarn start` and `yarn watch`?

Isn't this duplicating how we serve assets? Don't really see a need for node running it's own server.

If we decide to move this way though, it might make sense for us to server side render with a small express server. That way we can hydrate and make first page load way faster (we don't have to do it rn, we can do it whenever, even next term). See: https://reactjs.org/docs/react-dom-server.html

AbhiPrasad: +1 on moving it to another service tho

docker/Dockerfile.dev
```diff
@@ -1,16 +1,25 @@
...
 # Make sure we don't pollute the main docker-file
 # CompileDaemon allows for hot reloading
 RUN cd /tmp && go get github.com/githubnemo/CompileDaemon
 RUN go build
 
 EXPOSE 8900
-# Want to run yarn watch in the background for hot reloading
-CMD (cd ui && yarn watch &) && CompileDaemon --command="./shoe-project server"
+
+CMD go run ./shoe-project server
```
AbhiPrasad: let's go with exec form if it's a single command, it has better optimizations + works better with signals iirc.

```suggestion
CMD ["go", "run", "./shoe-project", "server"]
```

AbhiPrasad (commented):
Will approve after other PR is merged in.

docker-compose.yml
dhillondeep: yeah i did not realize it was `org` lol will fix

docker-compose.yml
dhillondeep: I made `frontend` run its own node server because this service can run independently and frontend developers don't need go server to run. When we are running all services, we can just ignore that node server

docker/Dockerfile.dev
dhillondeep: right! looks much cleaner

AbhiPrasad (approved):

PR #61, opened 2020-10-17
Server-side auth

### LINKED ISSUE #58: Create Authentication System
To Do:
- /login endpoint to get user details and verify hashed password against the one stored in the database
- on successful verification, create time scoped JWT token and send it back to the user
- update all the create endpoints to authenticate against JWT token provided

s278gupt:
Closes https://github.com/uwblueprint/shoe-project/issues/58

Everything server-side for now, no client storage (e.g. expect the token to be manually set as header, see usage below). This means no logout/refresh flow (token auto expires, that's it)

*Usage:*
```bash
➜  shoe-project git:(auth) curl http://localhost:8900/api/<protected-endpoint>                                                                                                     
> Unauthorized

➜ curl --header "Content-Type: application/json" \
  --request POST \
  --data '{"username":"<username>","password":"<password>"}' \
  http://localhost:8900/api/login
> {"status":"OK","payload":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VybmFtZSI6ImFkbWluIiwiZXhwIjoxNjAyOTQxMzEyfQ.A66yES7LtlRWYwCd0APwz3zfCYwmEVCQ2pwlPc51nTw"}

➜  shoe-project git:(auth) TOKEN="eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VybmFtZSI6ImFkbWluIiwiZXhwIjoxNjAyOTQxMzEyfQ.A66yES7LtlRWYwCd0APwz3zfCYwmEVCQ2pwlPc51nTw"

➜  curl -H 'Accept: application/json' -H "Authorization: Bearer ${TOKEN}" http://localhost:8900/api/<protected-endpoint>
```


config/defaults.go
```diff
@@ -37,4 +37,9 @@ func init() {
 	viper.SetDefault("database.name", "test")
 	viper.SetDefault("database.port", "5432")
 	viper.SetDefault("database.host", "localhost")
+
+	// Auth Defaults
+	viper.SetDefault("jwtKey", "GotFKl1PGgMpLg7D36NiI0hy/gsl6woTCXYdhKATbzc=")
```
AbhiPrasad: Not a fan of storing defaults in code, we can change this after though

internal/database/migrations/migrations.go
```diff
@@ -1,13 +1,27 @@
...
+	"golang.org/x/crypto/bcrypt"
 )
 
 func CreateTables(db *gorm.DB) error {
-	return db.AutoMigrate(&models.Story{}, &models.Author{})
+	return db.AutoMigrate(&models.Author{}, &models.Story{}, &models.User{})
+}
+
+func CreateSuperUser(db *gorm.DB) error {
+	// Clear current superuser if any
```
AbhiPrasad, line 16: What do you think about having multiple super users?

Honestly might be worth out to flesh out a basic user table/permissions rn that the admin panel will use.

AbhiPrasad (commented):
I’ll do a proper review on Sunday, this was just a first look (on mobile too lol). At initial glance good stuff! 

I’m of the opinion that we should build a users table, instead of just hard coding a super user and token - it’ll make things much easier, but I can be convinced otherwise.

I also think it makes sense for us to either a) extend the token much longer than 15 min 
b) every authenticated request, we keep updating the tokens expiry duration (aka the ping login, get the token, wait like 5 min, ping stories, and token is back to 15 min) 

I lean toward option A, cause it’s easier lol

config/defaults.go
dhillondeep: ^^ I second this! Since this is an environment variable anyway, we can set it up in docker-compose or something. We should not have `jwt` token in code

internal/database/migrations/migrations.go
```diff
@@ -1,13 +1,27 @@
...
+	db.Where("username = ?", "admin").Delete(models.User{})
+	
+	hashedPassword, _ := bcrypt.GenerateFromPassword([]byte(config.GetSuperUserPassword()), 8);
```
dhillondeep: I would recommend setting cost for `GenerateFromPassword` to be 10 based on the projects I have worked on and some research I have done in past.

restapi/api.go
```diff
@@ -23,9 +25,20 @@ func Router(db *gorm.DB) (http.Handler, error) {
 		logger:   zap.S(),
 	}
 
-	rest.GetHandler(r, "/health", api.health)
-	rest.GetHandler(r, "/stories", api.ReturnAllStories)
-	rest.GetHandler(r, "/story/{storyID}", api.ReturnStoryByID)
+	// Public API
+	r.Group(func(r chi.Router) {
+		rest.GetHandler(r, "/health", api.health)
+		rest.GetHandler(r, "/login/{username}/{password}", api.Login)
```
dhillondeep: I don't like how password is a part of URL when making GET request for `/login`. Can we make this a POST endpoint and pass username and password as a part of header?

restapi/api.go
```diff
@@ -23,9 +25,20 @@ func Router(db *gorm.DB) (http.Handler, error) {
...
+		rest.GetHandler(r, "/stories", api.ReturnAllStories)
+		rest.GetHandler(r, "/story/{storyID}", api.ReturnStoryByID)
+	})
+	
+	// Private API
+	r.Group(func(r chi.Router) {
+		r.Use(middleware.VerifyToken)
```
dhillondeep: `chi` has a middeware for this already. I think we should try using that instead of creating our own!

https://github.com/go-chi/jwtauth

restapi/auth.go
```diff
@@ -0,0 +1,54 @@
...
+		return rest.ErrInternal(api.logger, err)
+	}
+
+	// Validate password
+	if err = bcrypt.CompareHashAndPassword([]byte(user.Password), []byte(password)); err != nil {
+		return rest.ErrUnauthorized(fmt.Sprintf("Wrong password for username %s", username))
+	}
+
+	// Create JWT token with 15 minutes expiry for user
+	expirationTime := time.Now().Add(15 * time.Minute)
```
dhillondeep: can we also make `expirationTime` a config variable (environment variable)? This will allow us to have different values when in prod and when developing

restapi/auth.go
```diff
@@ -0,0 +1,54 @@
...
+	claim := &models.Claims {
+		Username: username,
+		StandardClaims: jwt.StandardClaims {
+			ExpiresAt: expirationTime.Unix(),
```
dhillondeep: Can you also add `Issuer` field to this? This can help with security

internal/database/migrations/migrations.go
dhillondeep: You are right it is worth having conversation about this. Maybe let's do that tomorrow!

dhillondeep (changes requested):

restapi/api.go
s278gupt: 🤦‍♀️ will swap out

config/defaults.go
s278gupt: hm wasn't sure where to put it - do you think we should also move the db stuff out of there?

s278gupt:
> we should build a users table, instead of just hard coding a super user and token

hmm not exactly sure what you mean. there is a users table rn, just that we're seeding it with a "superuser". do you mean we should look into user roles, or are you talking about a database session token (persisting a token for an authenticated user in the db)?

s278gupt:
> we should build a users table, instead of just hard coding a super user and token

hmm not exactly sure what you mean. there is a users table rn, just that we're seeding it with a "superuser". do you mean we should look into user roles, or are you talking about a database session token (persisting a token for an authenticated user in the db)?

s278gupt:
@AbhiPrasad 
> we should build a users table, instead of just hard coding a super user and token

hmm not exactly sure what you mean. this pr does create a users table, and also seeds it with a "superuser". did you mean we should look into user role-based access, or are you talking about a database session token (persisting a token for an authenticated user in the db)?

Assuming you're talking about user roles (authorization), I think that's separate from the basic auth(entication) in this pr. regardless of roles, the token creation/verification process in this pr remains the same (unless you're thinking of storing user roles in the jwt claim, which imo is not best practice). the authorization of what a user can/cannot do comes _after_ this basic auth. 

for example, we can use `jwt` for user authentication (login/signup/logout) and `Casbin` for managing roles and routes. Maybe this quick tutorial will show what I mean: https://zupzup.org/casbin-http-role-auth/. In this example, the additional changes to this pr may be to migrate the users table to include a `roles` column, and set policies using Casbin.

maybe I missed the point? (and sorry for not waiting until tomorrow, i prefer to have some time to think before we discuss irl 😛  )

config/defaults.go
dhillondeep: yeah I think DB stuff should also not be in code! we should talk about this sunday meeting

dhillondeep (approved):
🚢 🚢

PR #63, opened 2020-10-18
Add basic map to FE

### LINKED ISSUE #54: Add basic map to page
Part of #34

Let's use leaflet (https://github.com/PaulLeCam/react-leaflet) and open street maps.

lesliexin:
**Summary**
- Closes #54 
- Added leaflet and react-leaflet to project deps 
- Added a mapbox map centered at the university of Waterloo on the Shoe Map page

image

Referenced: 
https://medium.com/javascript-in-plain-english/leaflet-map-with-react-part-i-4ef4ecbdcc1b
https://stackoverflow.com/questions/37166172/mapbox-tiles-and-leafletjs


ui/src/pages/ShoeMap.tsx
```diff
@@ -1,11 +1,37 @@
...
   return (
-    <div>
-      <TitleText>ShoeMap</TitleText>
-      <span> test text</span>
-    </div>
+    <MapContainer>
+      <TitleText>Shoe Map</TitleText>
+      <StyledMap center={currentLocation} zoom={zoom}>
+        <TileLayer
+          url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
```
aspensantos: I think we discussed using mapbox instead of openstreetmap, since even though mapbox is not open source it will be free for our usage level.

ui/src/pages/ShoeMap.tsx
```diff
@@ -1,11 +1,37 @@
...
+  height: 100%;
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+`;
+
 export const ShoeMap: React.FC = () => {
+  const zoom = 12;
+  const currentLocation = { lat: 43.4723, lng: -80.5449 };
```
aspensantos, line 23: Should these be props? If we are going to reuse this component for maps in the stories that could be helpful but if not it is probably fine like this.

ui/src/pages/ShoeMap.tsx
lesliexin: Ooh okay I'll change it

ui/src/pages/ShoeMap.tsx
lesliexin: Yeah for sure, I just hardcoded the values temporarily. I think I'll hold off on adding them to the props until we have a clearer idea of what we're building

ui/package.json
```diff
@@ -25,9 +25,13 @@
     "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^11.1.0",
     "@types/jest": "^26.0.14",
+    "@types/leaflet": "^1.5.19",
```
AbhiPrasad, line 28: We can do this another PR, but we should move the `@types` dependencies to `devDependencies`

ui/src/pages/ShoeMap.tsx
```diff
@@ -1,11 +1,35 @@
 import * as React from "react";
+import styled from "styled-components";
+import { Map, TileLayer } from "react-leaflet";
+import "leaflet/dist/leaflet.css";
+
 import { TitleText } from "../styles/typography";
 
+const StyledMap = styled(Map)`
+  height: 70vh;
```
AbhiPrasad, line 9: Took `vh`? from the examples?

ui/src/pages/ShoeMap.tsx
```diff
@@ -1,11 +1,35 @@
 import * as React from "react";
+import styled from "styled-components";
+import { Map, TileLayer } from "react-leaflet";
+import "leaflet/dist/leaflet.css";
```
AbhiPrasad, line 4: We should move all the css dist files to a common file and import them into `index.tsx`.

AbhiPrasad (approved):
Think we should be good after rebase.

ui/src/pages/ShoeMap.tsx
lesliexin: This file is imported from the leaflet package. Could you elaborate?

ui/src/pages/ShoeMap.tsx
lesliexin: No, I just added that for now

ui/src/pages/ShoeMap.tsx
AbhiPrasad: You know what, after looking at this more, I think it's fine as is, ignore my point lol.

PR #64, opened 2020-10-18
feat: Add data fetching to UI

### LINKED ISSUE #53: Add data fetching to UI
We should add some basic data fetching to the app.

Options:
- plain `window.fetch` (`fetch` doesn't work with IE11 so you'll have to have https://github.com/matthew-andrews/isomorphic-fetch)
- axios
- XMLHTTPRequest 😎 

I really like using https://github.com/vercel/swr

AbhiPrasad:
Resolves #53 

Added https://github.com/vercel/swr for easy data fetching.

For those of you unfamiliar, `swr` is a data fetching framework built with react hooks. It's made by vercel, the guys who also make `next.js`. SWR handles queries/mutations, interval polling, cache revalidation and invalidation and works with react suspense (we're not gonna use it, but maybe in the future?)

Just look at all these fancy options! https://github.com/vercel/swr#options

It even gives some great patterns for sending `POST` requests and but eagerly updating the UI https://github.com/vercel/swr#mutation-and-post-request (meaning you can write less state stuff to make it happen).

It's low byte size, uses native `window.fetch` (but allows you to use any fetcher, like XHR too), and I give it my seal of approval 👍 

I'll add test soon tm;

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,5 @@ clean-backend:
 
 clean-docker:
 	docker-compose down
-	docker image rm shoe-project_server || true
 
 clean: clean-docker clean-frontend clean-backend
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3.7'
+version: "3.7"
 
 services:
   build-setup:
@@ -6,8 +6,8 @@ services:
     container_name: shoe_project_build_setup
     build:
       context: .
-      dockerfile: './docker/Dockerfile.dev'
-    command: './shoe-project seed'
+      dockerfile: "./docker/Dockerfile.dev"
+    command: "./shoe-project seed"
     env_file:
       - .env
     depends_on:
@@ -17,14 +17,13 @@ services:
     image: shoe_project_image
     container_name: shoe_project_server
     ports:
-      - '8900:8900'
+      - "8900:8900"
     command:
       - /bin/sh
       - -c
-      - |
-          CompileDaemon --command="./shoe-project server"
+      - CompileDaemon --command="./shoe-project server"
     volumes:
-      - './:/app'
+      - "./:/app"
     env_file:
       - .env
     depends_on:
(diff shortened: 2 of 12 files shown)


Makefile
```diff
@@ -27,6 +27,5 @@ clean-backend:
 
 clean-docker:
 	docker-compose down
-	docker image rm shoe-project_server || true
```
AbhiPrasad, line 31: Removed cause doesn't apply anymore

docker-compose.yml
```diff
@@ -1,14 +1,14 @@
-version: '3.7'
+version: "3.7"
```
AbhiPrasad, line 1: This is my IDE formatting, I can revert if necessary.

ui/src/App.tsx
```diff
@@ -1,31 +1,46 @@
...
+import { SWRConfig } from "swr";
 import { BrowserRouter as Router, Switch, Route, Link } from "react-router-dom";
 
 import { Home, ShoeMap } from "./pages";
 
+const defaultFetcher = (
+  input: RequestInfo,
+  init?: RequestInit
+): Promise<unknown> =>
+  fetch(input, init)
```
AbhiPrasad, line 11: Just using native fetch here, but we can use whatever people want.

ui/src/pages/Home.tsx
```diff
@@ -1,24 +1,23 @@
...
+
   return (
     <div>
-      <TitleText>Home</TitleText>
-      <span> test text hehe</span>
-      <Button>Test Button</Button>
+      <TitleText>Stories</TitleText>
+      {data.map((story) => (
+        <div key={story.ID}>
+          <div>{story.Title}</div>
```
AbhiPrasad: These have to be uppercase for now as that's what the server sends, will change after server side change is made.

ui/src/App.tsx
```diff
@@ -1,31 +1,47 @@
...
-          </li>
-          <li>
-            <Link to="/map">ShoeMap</Link>
-          </li>
-        </ul>
-      </nav>
+    <SWRConfig
+      value={{
+        fetcher: defaultFetcher,
+        dedupingInterval: 0,
```
aspensantos: I am not sure we should remove deduping by setting this to 0, our stories and authors should not change that frequently, so reducing duplicate requests for the sake of efficiency should be fine.

AbhiPrasad: Yup, good catch put that in there for testing. I can change.

aspensantos (approved):

PR #65, opened 2020-10-18
Fix docker-compose pulling image issue

dhillondeep:
There was an issue regarding image not found so this is a fix for that

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ default: docker-run
 
 setup:
 	(cd ui; yarn)
+	docker-compose build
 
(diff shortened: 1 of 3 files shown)


AbhiPrasad (approved):

Makefile
```diff
@@ -2,6 +2,7 @@ default: docker-run
 
 setup:
 	(cd ui; yarn)
+	docker-compose build
```
hanlinc27, line 6: bit of an unrelated note but the README should be` make docker-database `not `make docker-postgres` I think?

PR #73, opened 2020-10-22
feat: Add design system colors and font

### LINKED ISSUE #72: Bring Colors and Theme from design to frontend

AbhiPrasad:
Resolves #72 

Based on the designs from Figma.

- Added the Poppins font for the body
- Added all the colours we should be using
- Couldn't added the `Canela` font because it we can't obtain it for free. I will open another issue for that specifically after this one merges in (either figure out how to get it, or use an alternative)


ui/src/index.css
```diff
@@ -0,0 +1,7 @@
+body {
+  font-family: Poppins;
+  font-style: normal;
+  font-weight: normal;
+  font-size: 14px;
+  line-height: 24px;
```
lesliexin, line 6: Do we need to specify these?

lesliexin (approved):
Looks good!

ui/src/index.css
AbhiPrasad: I just grabbed them off the figma css, I'll keep em for now, but we can always change it later.

PR #75, opened 2020-10-22
Add linting, checks, and tests locally and github actions

dhillondeep:
- Expand Makefile to add `backend-fmt` and `backend-check` commands
- Developer can format and check backend code before pushing
- Added linting checks to Github actions
- Split actions into jobs and added tests as a job


dhillondeep:
Go lint will fail but we can fix that in upcoming PR

AbhiPrasad (approved):

PR #78, opened 2020-10-25
City to coords

aspensantos:


restapi/authors.go
```diff
@@ -2,23 +2,47 @@ package restapi
...
 
 	// respond to the client with the error message and a 400 status code.
-	if err := json.NewDecoder(r.Body).Decode(&author); err != nil {
+	if err := json.NewDecoder(r.Body).Decode(&authors); err != nil {
 		return rest.ErrInvalidRequest(api.logger, "Invalid payload", err)
 	}
 
-	if err := api.database.Create(&author).Error; err != nil {
+	token := config.GetMapBoxToken()
+	mapBox, err := mapbox.NewMapbox(token)
```
AbhiPrasad: Can we init mapbox earlier so it is available to every post handler?

restapi/authors.go
```diff
@@ -2,23 +2,47 @@ package restapi
...
+	forwardOpts.Limit = 1
+	forwardOpts.Country = "CA"
+
+	for i := 0; i < len(authors); i++ {
+		place := authors[i].CurrentCity
+		forward, err := mapBox.Geocode.Forward(place, &forwardOpts)
+		if err != nil {
+			return rest.ErrInvalidRequest(api.logger, fmt.Sprintf("Story %d is has an invalid current city", i), err)
+		}
+		authors[i].Latitude = float32(forward.Features[0].Center[0])
```
AbhiPrasad: Should we just change our types to match the mapbox latitude and longitude types?

restapi/authors.go
```diff
@@ -2,23 +2,47 @@ package restapi
...
-	if err := api.database.Create(&author).Error; err != nil {
+	token := config.GetMapBoxToken()
+	mapBox, err := mapbox.NewMapbox(token)
+	if err != nil {
+		return rest.ErrInvalidRequest(api.logger, "Invalid mapbox token", err)
+	}
+
+	var forwardOpts geocode.ForwardRequestOpts
+	forwardOpts.Limit = 1
+	forwardOpts.Country = "CA"
```
AbhiPrasad: lets make `forwardOpts` a constant outside of the func, no need to redeclare every func call.

AbhiPrasad (changes requested):
Seems 100000 requests -> free, which seems super reasonable to me (https://www.mapbox.com/pricing/)

If we want to save even more on the requests, we use something like redis and cache (and bust the cache like every month or 2), but I think thats optional (unless someone really wants to learn how to do it)

restapi/authors.go
aspensantos: Since go doesn't support constant structs I made constants for the limit and country instead and kept those outside the func.

AbhiPrasad:
I had the chance to think about this a little bit more.

Can we potentially download a csv off https://www.nrcan.gc.ca/earth-sciences/geography/download-geographical-names-data/9245, then filter that table down to cities, and then just use that list? 

It'll be easy for us to query too when we build the admin center and people have to enter a valid city name.

AbhiPrasad (approved):

dhillondeep:
@m-penny so all looks good but currently there is no way to mock this API for testing. So i will add some code to make it easier to test!

dhillondeep:
I 🎩 the code to make sure it all works!

AbhiPrasad (approved):

aspensantos:
> @m-penny so all looks good but currently there is no way to mock this API for testing. So i will add some code to make it easier to test!

LGTM! There are some conflicts in go.sum and go.mod but I am not sure how to resolve them. Can someone who has permission merge this? @dhillondeep  @AbhiPrasad

dhillondeep:
@m-penny yeah I am working on that right now! Going to get this merged soon

PR #79, opened 2020-10-28
Adding pin preview component

lesliexin:
Added Pin Preview component using[ leaflet popup](https://leafletjs.com/reference-1.7.1.html#popup), styled with MaterialUI.
pin-preview


ui/package.json
```diff
@@ -22,6 +22,8 @@
     "test": "jest"
   },
   "dependencies": {
+    "@material-ui/core": "^4.11.0",
+    "@material-ui/types": "^5.1.0",
```
AbhiPrasad: Why are we pulling in `@material-ui/types`?

AbhiPrasad (approved):
Maybe it makes sense to add storybook so we can easily test components

ui/src/components/PinPreview.tsx
```diff
@@ -1,41 +1,64 @@
...
+  author,
+  date,
+  country,
+}: PinPreviewProps): JSX.Element {
 
   return (
     <StyledCard>
-      <StyledMedia/>
+      <StyledMedia 
+        image={require('../assets/images/temp.png')}
```
AbhiPrasad: Don't think we need `require()` here, you should just be able to do `import image from "../assets/images/temp.png"` and supply that to the prop.

We shouldnt be using `commonJS` require imports in this code base (we can remove the node types and typescript changes then too)

AbhiPrasad:
You'll probably need to rebase as we merged in #82 (adds material UI in that PR)

ui/src/components/PinPreview.tsx
lesliexin: Yeah Hanlin and I wanted some help with this. TS wasn't letting us directly import nor pass the path to the prop. We added the require as a temp working solution

ui/.tsconfig.json
```diff
@@ -11,7 +11,9 @@
         "allowJs": true,
         "jsx": "react",
         "esModuleInterop": true,
-        "allowSyntheticDefaultImports": true
+        "allowSyntheticDefaultImports": true,
+        "types": ["node"],
+        "typeRoots": [ "../node_modules/@types" ]
```
AbhiPrasad: Can we remove this?

ui/custom.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module "*.svg" {
+    const content: any;
+    export default content;
+  }
+declare module "*.png" {
```
AbhiPrasad: Prettier didn't run on these files?

ui/package.json
AbhiPrasad: ^ Bump to get this addressed before we merge.

ui/src/components/Pin.tsx
```diff
@@ -23,12 +45,26 @@ export function Pin({
     iconUrl: state === PinState.Unfocused ? unfocused : resting,
     iconRetinaUrl: state === PinState.Unfocused ? unfocused : resting,
     iconAnchor: state === PinState.Selected ? [29, 70] : [25, 61],
-    popupAnchor: null,
+    popupAnchor: [-234, 200],
```
AbhiPrasad, line 48: Any reason we did these numbers in particular?

ui/.tsconfig.json
AbhiPrasad: Can we remove this?

ui/custom.d.ts
AbhiPrasad: We'll need to eslint-disable the any here too, just search the codebase for `// eslint-disable-next-line` to see how it's done

ui/src/components/index.ts
```diff
@@ -0,0 +1 @@
+export { PinPreview } from "./PinPreview";
```
AbhiPrasad, line 2: Might be prudent to add the Pin component here too since you made the index already.

ui/custom.d.ts
AbhiPrasad: Oh ok there seems to be two declaration files, this one `ui/custom.d.ts`, and then `ui/src/global.d.ts`.

Let's get rid of one and choose the other (im in favour of `global.d.ts`)

AbhiPrasad (changes requested):
After comments addressed just comment on the PR and ill approve + merge

ui/custom.d.ts
hanlinc27: Going to keep the `ui/custom.d.ts` (+rename it) and delete the other custom types file. `global.d.ts` in /src doesn't seem to be resolving the svg module errors.

ui/src/components/Pin.tsx
hanlinc27: So it's bit rough, but popup Anchor is basically the offset from the pin --> horizontal offset = half of popup to middle of pin and vertical offset is half of vertical popup-half of popup (vertically). Took coordinates off Figma and also eye-balled it a bit.

AbhiPrasad (approved):

ui/package.json
lesliexin: TIL that material ui has its own types!! Will remove

PR #82, opened 2020-10-29
Pin Component

### LINKED ISSUE #69: Create React Pin Component

Shehryar21:
Resolves #69


AbhiPrasad (approved):
Mind running `go mod tidy` to clean up the `go.mod` changes?

Please address the eslint issues, you can run `yarn fix` in the ui folder, otherwise LGTM.

ui/src/svg.d.ts
```diff
@@ -0,0 +1,5 @@
+// svg-shim.d.ts
+declare module "*.svg" {
+  const content: any;
```
AbhiPrasad, line 4: You can do `// eslint-disable-next-line @typescript-eslint/no-explicit-any` to ignore this `Unexpected any. Specify a different type ` warning.

We lint against this because we typically don't want people to use `any`, we want stronger types than that, but in this case we should make the svg modules as generic as possible.

ui/src/components/Pin.tsx
```diff
@@ -0,0 +1,31 @@
...
+  Selected,
+  Unfocused,
+}
+
+export interface PinProps {
+  state?: PinState;
+  position: [number, number];
+}
+
+export function Pin({ state = PinState.Resting, position }: PinProps) {
```
AbhiPrasad: Can you add `JSX.Element` as a return type here?

You can search the codebase for `JSX.Element` to see how it is used.

ui/src/pages/ShoeMap.tsx
```diff
@@ -3,6 +3,7 @@ import styled from "styled-components";
 import { Map, TileLayer } from "react-leaflet";
 import "leaflet/dist/leaflet.css";
 
+import { Pin, PinState } from "../components/Pin";
```
hanlinc27, line 6: Delete unused import

hanlinc27 (approved):

hanlinc27:
this looks really good! 😄 It'll be good to build the popup preview from this as well. A few small questions: 
- Is there a reason there's no redirect state at the moment?
- Why is the unfocused pin larger than the focused pin? Seems like it should be the opposite on Figma? (Figma)
image
(feat/react-pin branch)
image

- I had to add a custom types file to resolve the svg imports 
`declare module "*.svg" {
  const content: any;
  export default content;
}`
Edit: nevermind, saw you already did that

ui/src/pages/ShoeMap.tsx
Shehryar21: we have to use it to render pin that have the unfocused and selected state, the default state is resting.

Shehryar21:
Idk for some reason the unfocused pin appears to be bigger than the resting pin. I checked the size of both of them and its the same (50 x 61).

hanlinc27:
@Shehryar21 A quick fix would be to set unfocused.svg to have `<svg width="58" height="66" viewBox="0 0 58 66"` to be consistent with that of resting.svg. I'm not sure why but when you set `L.Point(50,61)` it constrains/sets the icon for resting to that size but unfocused.svg already has those dimensions and it's not constraining it properly?

Seems cleared up then:
image
Otherwise, everything else looks fine 🙂

Shehryar21:
@hanlinc27 yeah thats weird but it works. I changed the size of unfocused.svg and also added an iconAnchor so that if you zoom out in the map, the pins hold their position.

PR #83, opened 2020-10-30
Endpoints Testing

### LINKED ISSUE #51: Create api integration tests

hanlinc27:
Compilation of tests for: 

- GET stories/stories by id 
- POST login/create author/create stories
- Health Check 

References to:
[Testify](https://github.com/stretchr/testify)
[httpexpect](https://github.com/gavv/httpexpect/blob/master/_examples/fruits.go)
[gorm sqlite](https://gorm.io/docs/connecting_to_the_database.html) for mocking DB

To run tests:
`cd restapi`
`go test -v`

image


restapi/endpoints_test.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+}
+
+func (suite *endpointTestSuite) TestHealthCheck() {
+	suite.endpoint.GET("/health").
+		Expect().
+		Status(http.StatusOK).JSON().Object().Value("message").String().Equal("Hello World")
+}
+
+func (suite *endpointTestSuite) TestGetAllStories() {
```
s278gupt, line 75: shouldn't we add a few stories in the db and then test this call?

restapi/endpoints_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (suite *endpointTestSuite) SetupSuite() {
+	db, err := testutils.MockDatabase()
+	if err != nil {
+		suite.Fail("error while creating database", err)
+	}
+	suite.db = db
+
+	// setup required for jwt authentication
+	viper.SetDefault("auth.jwt_key", "GotFKl1PGgMpLg7D36NiI0hy/gsl6woTCXYdhKATbzc=")
```
s278gupt: please make sure that our secrets env vars are different in the `.env` file. feel free to generate new tokens, users, passwords etc and send us the updated `.env` file.

restapi/endpoints_test.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	suite.endpoint.POST("/stories").WithHeader("Authorization", "Bearer "+suite.token).
+	WithJSON(json).
+	Expect().
+	Status(http.StatusOK)
+
+	var storyCount int64
+	suite.db.Table("stories").Count(&storyCount)
+	suite.Equal(1, int(storyCount))
```
s278gupt, line 164: Is it worth it to test if the actual record inserted in the db has the same field values as our json payload?

restapi/endpoints_test.go
```diff
@@ -0,0 +1,146 @@
...
+func (suite *endpointTestSuite) SetupSuite() {
+	db, err := testutils.MockDatabase()
+	if err != nil {
+		suite.Fail("error while creating database", err)
+	}
+	suite.db = db
+
+	// setup required for jwt authentication
+	viper.SetDefault("auth.jwt_key", "GotFKl1PGgMpLg7D36NiI0hy/gsl6woTCXYdhKATbzc=")
+	viper.SetDefault("auth.jwt_expiry", "2h")
```
s278gupt: nit: if we're logging in on every test setup, why do we need the token expiry to be this long?

restapi/endpoints_test.go
```diff
@@ -0,0 +1,146 @@
...
+		},
+	}
+
+	suite.endpoint.POST("/stories").WithHeader("Authorization", "Bearer "+suite.token).
+	WithJSON(json).
+	Expect().
+	Status(http.StatusOK)
+	suite.endpoint.GET("/story/1").
+		Expect().
+		Status(http.StatusOK)
```
s278gupt: Do we get a payload with some story attributes? We should probably check they're equal to the expected story?

testutils/database.go
```diff
@@ -0,0 +1,35 @@
...
+func CloseDatabase(db *gorm.DB) error {
+	sqlDB, err := db.DB()
+	if err != nil {
+		return err
+	}
+
+	return sqlDB.Close()
+}
+
+func DropTables(db *gorm.DB) error {
```
s278gupt, line 26: drop `users` table as well?

restapi/endpoints_test.go
hanlinc27: Hm yeah you're right, could also just keep it as 2h to be consistent?

restapi/endpoints_test.go
hanlinc27: yup that make sense! Will throw in a few stories 👍

restapi/endpoints_test.go
hanlinc27: I don't think it actually matters what I set the` auth.jwt_key` here? Setting it as `	viper.SetDefault("auth.jwt_key", "random")
` still passes all the tests? 😕

restapi/endpoints_test.go
hanlinc27: I don't think so, since we already test that it increments the story count by 1 successfully. We also know that it does create the correct stories since we now verify correct payload from **POST stories** in **GET all stories** 🙂

restapi/endpoints_test.go
hanlinc27: yes, adding that in!

restapi/endpoints_test.go
s278gupt: yes, that's the way it should be in all passing tests (because we are making sure to send the correct token in the header for all our post requests).

actually, this makes me think we should have failing tests for login (i.e. username or pw is incorrect), and tests for unauthorized POST requests (where the token is missing or incorrect). _but_, because our app is quite small, we maybe can get away with testing only happy paths. i don't have a strong opinion, maybe others can comment

restapi/endpoints_test.go
s278gupt: it looks like we're dropping tables in the teardown method after each test, so we can't have a separate GET test to verify that this post was correct. regardless, i don't think we should be using a GET request to test the story was inserted into the db correctly, because that relies on the `get story by id` endpoint working correctly. instead, if we are checking the record in the db, we should do so directly by querying the db.

restapi/endpoints_test.go
s278gupt: consistent with?

restapi/endpoints_test.go
s278gupt: i think we should be adding the stories directly in the mock db rather than using POST requests.

restapi/endpoints_test.go
hanlinc27: Since the real token expiry is 2h?

restapi/endpoints_test.go
dhillondeep: ^^agree

restapi/endpoints_test.go
dhillondeep: we should have a different suite all together for login testing

restapi/endpoints_test.go
dhillondeep: yeah maybe set it as `5m` that should be fine

restapi/endpoints_test.go
hanlinc27: ok fixed it to add directly to db!

restapi/endpoints_test.go
hanlinc27: @s278gupt we discussed in the wed meeting to put login testing and get countries testing as a separate ticket ^^

hanlinc27:
@s278gupt @dhillondeep Should be good to go now?

dhillondeep:
I will rebase and resolve conflicts

restapi/endpoints_test.go
```diff
@@ -0,0 +1,197 @@
...
+
+func (suite *endpointTestSuite) SetupTest() {
+	if err := migrations.CreateTables(suite.db); err != nil {
+		suite.Fail("error while creating tables", err)
+	}
+	if err := migrations.CreateSuperUser(suite.db); err != nil {
+		suite.Fail("error creating super user", err)
+	}
+
+	suite.token = suite.endpoint.POST("/login").
```
s278gupt, line 54: sorry for nit :/ 

i think this should be inserted directly into db too, instead of posting

restapi/endpoints_test.go
```diff
@@ -0,0 +1,197 @@
...
+
+	json_story2 := []models.Story{
+		{
+			Title:    "Hitchhiker's Guide to the Galaxy",
+			Content:  "Fiction",
+			AuthorID: 2,
+		},
+	}
+
+	//Directly add to mock DB
```
s278gupt: side note: we're not mocking the db anymore, more like a sleek/light db (using sqlite)

so really small nit: we should remove references to "mock db" like in this comment and the `mock` var name below (unless thats convention, im not really sure)

restapi/endpoints_test.go
```diff
@@ -0,0 +1,197 @@
...
+	suite.db.Create(&json_story1)
+	suite.db.Create(&json_story2)
+
+	var response = suite.endpoint.GET("/stories").
+		Expect().
+		Status(http.StatusOK).
+		Status(http.StatusOK).JSON()
```
s278gupt, line 98: just curious why the two `Status(http.StatusOk)`?

restapi/endpoints_test.go
hanlinc27: Don't we need a POST so we can grab the response?

restapi/endpoints_test.go
hanlinc27: oops my bad, will take out repeat

restapi/endpoints_test.go
```diff
@@ -0,0 +1,197 @@
...
+			"author_id": 2,
+			"content": "Fiction",
+			"title": "Hitchhiker's Guide to the Galaxy"
+		}
+		],
+		"status": "OK"
+	}`
+
+	//Verify mock matches response
+	response.Schema(mock)
```
s278gupt, line 117: i was curious to see whether `.Schema` only validated the json schema (i.e. the shape of the json), so I tried to run the test with the following json:

```go
	mock := `{
		"payload": [
		{
			"author_id": 4,
			"content": "blablablabla",
			"title": "yes"
		},
		{
			"author_id": 5,
			"content": "",
			"title": "i dont exist"
		}
		],
		"status": "OK"
	}`
```

...and it still worked :( 

so `response.Schema(...)` only checks the shape of the response, but not the actual values. i still think it's valuable to check the actual values of the stories in the response - this would involve doing something like:

```
response.Value("payload").Array().Element(0).Object().Value("ID").Equal(1)
```

@hanlinc27  @dhillondeep what do u think?

restapi/endpoints_test.go
s278gupt: yes yes!!!

you're right, my bad.

s278gupt (approved):
lgtm, just one comment about checking the response values vs schema

restapi/endpoints_test.go
hanlinc27: added that in :) should be all good now 👍

PR #86, opened 2020-11-01
Fix linter errors

### LINKED ISSUE #76: Fix linting issues in Go code

aspensantos:
Closes #76

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -79,7 +79,7 @@ var (
 			}
 
 			if err := zap.L().Sync(); err != nil { // flush the logger
-				logger.Fatalw("Could not flush logger", "Err", err)
+				fmt.Fprint(os.Stderr, "Could not flush logger", err.Error())
 			}
 
 		},
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -68,7 +68,7 @@ var (
 						logger.Fatalw("Server listen and serve", "Err", err)
 					}
 					break readChannel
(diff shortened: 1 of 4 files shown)


cmd/server.go
```diff
@@ -78,7 +78,9 @@ var (
 				}
 			}
 
-			zap.L().Sync() // flush the logger
+			if err := zap.L().Sync(); err != nil { // flush the logger
+				logger.Fatalw("Could not flush logger", "Err", err)
```
dhillondeep: Since we are dealing with logger closing, we should not print error using logger. Instead we should do:
```go
fmt.Fprint(os.Stderr, "Could not flush logger", err.Error())
```

restapi/rest/http.go
```diff
@@ -13,7 +13,9 @@ type routeHandler func(w http.ResponseWriter, r *http.Request) render.Renderer
 // Custom ServeHTTP that handlers custom render errors
 func (fn routeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if renderer := fn(w, r); renderer != nil {
-		render.Render(w, r, renderer)
+		if err := render.Render(w, r, renderer); err != nil {
+			http.Error(w, err.Error(), 500)
```
dhillondeep: use `http.StatusInternalServerError` instead of typing error code as a number

dhillondeep (changes requested):

dhillondeep (approved):

PR #96, opened 2020-11-05
feat: Update map routes

### LINKED ISSUE #92: Change routes to show map when web app is loaded
Currently the map is accessible at the `/map` route. This needs to be changed and the map needs to be accessible at the `/` route.

AbhiPrasad:
Resolves #92 

Moves the map to the index route.

Also add routing for the pins. Basically every time you open a pin preview, it updates the URL query params. That way someone can save their position on the map. I found this to be a good quality of life feature, but I can remove it (or we can keep it for now, and change it when we add the stories page).

https://streamable.com/2cejru


lesliexin (approved):
Tested locally and looks good. I think it's fine to include the unique pin paths in this PR and I see how it would be useful.

ui/src/pages/ShoeMap.tsx
```diff
@@ -19,21 +20,74 @@ const MapContainer = styled.div`
...
+    (id: string) => {
+      return () => {
+        setSelectedPin(id);
+        history.push({
+          search: `?location=${id}`,
+        });
+      };
+    },
+    [setSelectedPin]
+  );
```
lesliexin, line 63: maybe this is a noob question but do you know why we need the nested functions here?

AbhiPrasad: It's because its a curried function, check the how the prop is used by the `Pin` component.

Basically the Pin component calls the function with an id, and so we have to return a function that accepts an id to be used by the click handler.

lesliexin: Ooh ya I didn't catch that :) Thanks!

ui/src/components/Pin.tsx
```diff
@@ -33,27 +33,59 @@ export enum PinState {
...
+// This is extremely jank
+// TODO: Figure out how anchors actually work
+function getAnchor(state: PinState): PointTuple {
+  switch (state) {
+    case PinState.Resting:
+      return [25, 61];
+    case PinState.Selected:
+      return [29, 70];
+    case PinState.Unfocused:
+      return [22, 57];
```
hanlinc27, line 52: Why is there a different iconAnchor for _unfocused_ compared to _resting_? If I understand it correctly, iconAnchor is where we would like to place the icon image. It defaults to the [lat,lng] we provide being the upper left corner of the svg and then we'd position it accordingly based on image size.

AbhiPrasad:
Closing because should be addressed by https://github.com/uwblueprint/shoe-project/pull/133 instead (I'm not trying to rebase lol)

PR #97, opened 2020-11-06
Inserting authors should be unique

s278gupt:
The set `(FirstName, LastName, OriginCountry)` identifies each author uniquely now. If any one of them is different, a new author is created. If all three are the same, a `duplicate primary key` error is thrown.

The foreign keys in `stories` are updated to these three fields. `AuthorID` is no longer column or foreign key in the `stories` table.

aside: I'm starting to think we should have unit tests for db.

### DIFF
--- a/internal/database/models/models.go
+++ b/internal/database/models/models.go
@@ -1,26 +1,33 @@
 package models
 
 import (
+	"time"
+
 	"gorm.io/gorm"
 )
 
 type Author struct {
-	gorm.Model
-	FirstName     string  `gorm:"not null" json:"first_name"`
-	LastName      string  `json:"last_name"`
-	Bio           string  `gorm:"type:text" json:"bio"`
-	OriginCountry string  `gorm:"not null" json:"origin_country"`
-	Stories       []Story `gorm:"foreignKey:AuthorID;constraint:OnUpdate:CASCADE,OnDelete:CASCADE;" json:"stories,omitempty"`
+	ID            uint `gorm:"autoIncrement:true"`
+	CreatedAt     time.Time
+	UpdatedAt     time.Time
+	DeletedAt     gorm.DeletedAt `gorm:"index"`
+	FirstName     string         `gorm:"primaryKey;not null" json:"first_name"`
+	LastName      string         `gorm:"primaryKey" json:"last_name"`
+	Bio           string         `gorm:"type:text" json:"bio"`
+	OriginCountry string         `gorm:"primaryKey;not null" json:"origin_country"`
+	Stories       []Story        `gorm:"foreignKey:AuthorFirstName,AuthorLastName,AuthorCountry;references:FirstName,LastName,OriginCountry;constraint:OnUpdate:CASCADE,OnDelete:CASCADE;" json:"stories,omitempty"`
 }
 
 type Story struct {
 	gorm.Model
-	Title       string  `gorm:"not null" json:"title"`
-	Content     string  `gorm:"type:text;not null" json:"content"`
-	CurrentCity string  `gorm:"not null" json:"current_city"`
-	Summary     string  `gorm:"type:text" json:"summary"`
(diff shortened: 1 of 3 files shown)


internal/database/models/models.go
```diff
@@ -2,24 +2,30 @@ package models
...
-	gorm.Model
-	FirstName     string  `gorm:"not null" json:"first_name"`
-	LastName      string  `json:"last_name"`
-	Bio           string  `gorm:"type:text" json:"bio"`
-	OriginCountry string  `gorm:"not null" json:"origin_country"`
-	CurrentCity   string  `gorm:"not null" json:"current_city"`
-	Latitude      float64 `gorm:"not null" json:"latitude"`
-	Longitude     float64 `gorm:"not null" json:"longitude"`
-	Stories       []Story `gorm:"foreignKey:AuthorID;constraint:OnUpdate:CASCADE,OnDelete:CASCADE;" json:"stories,omitempty"`
+	ID        		uint	`gorm:"autoIncrement:true"`
```
s278gupt, line 9: should we get rid of this? it's not being used anywhere right now, and I doubt we will need to query it in the future.

internal/database/models/models.go
```diff
@@ -2,24 +2,30 @@ package models
...
-	CurrentCity   string  `gorm:"not null" json:"current_city"`
-	Latitude      float64 `gorm:"not null" json:"latitude"`
-	Longitude     float64 `gorm:"not null" json:"longitude"`
-	Stories       []Story `gorm:"foreignKey:AuthorID;constraint:OnUpdate:CASCADE,OnDelete:CASCADE;" json:"stories,omitempty"`
+	ID        		uint	`gorm:"autoIncrement:true"`		
+	CreatedAt 		time.Time
+	UpdatedAt 		time.Time
+	DeletedAt 		gorm.DeletedAt `gorm:"index"`
+	FirstName     	string  `gorm:"primaryKey;not null" json:"first_name"`
+	LastName      	string  `gorm:"primaryKey" json:"last_name"`
```
AbhiPrasad: Multiple primary keys doesn't make sense, there should only be one (hence *primary* key).

I don't think it should be the database concern to check for uniqueness, that is an application concern (as fields and field types could easily change with a migration).

When we are create new authors, we can query and check for the existing authors. If one of the existing authors are good enough, we can select one of them, otherwise we create a new one. This also means we don't lose the possibility of people having the same firstname/lastname/origin.

s278gupt: yep, so we decided with @dhillondeep that the uniqueness of the author will be determined by the combo of their first/last/country, which is why we decided to make the three fields into a composite primary key. (we don't want the possibility of authors have the same first/last/country. for example, if author1 and author2 have the same firstname/lastname/country fields, we assume they're the same author. and i believe we have a separate issue to make `current_city` part of the `stories` model)

> I don't think it should be the database concern to check for uniqueness

imo, it's a good idea to make sure our primary keys are unique at the lowest level to ensure data consistency. right now (with this pr) we are just adding constraints for our db, i.e. data validation (not business logic). i don't think creating a composite primary key is equivalent to, for example, checking that the `origin_country` is indeed an actual country...as for adding a separate layer of validation by querying - well, we'd still want to filter based on first&last&country, so we'd prob want to have an index on those fields anyway...

did i understand your concern correctly?

internal/database/models/models.go
```diff
@@ -1,25 +1,31 @@
 package models
 
 import (
+	"time"
+
 	"gorm.io/gorm"
 )
 
 type Author struct {
```
s278gupt, line 9: removed `id` field from `Author`:

if we have `id` in Author, we have to set `auto_increment=True` (so that we don't have to manually pass in the `id` when inserting a new author). but for some reason, any integer column with `auto_increment` automatically becomes a primary key (and we don't want that). 

soo I just got rid of `id` column

i don't fully understand the magic that is happening here, does anyone know better?

dhillondeep (approved):

PR #98, opened 2020-11-08
Pin clustering on Zoom Out

### LINKED ISSUE #91: Collapse story pins when map is zoomed out (collapse)

hanlinc27:
Relevant links: 
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/leaflet.markercluster/index.d.ts
https://github.com/Leaflet/Leaflet.markercluster
Essentially we group which markers we would like to be clustered (i.e. the 10 from Toronto) and pass it in as an array to the cluster component. Alternatively, we could also pass in all the markers into the cluster and I could set different pin sizes based on the size of the cluster. Also was unsure if the marker clusters should exist on a separate map layer but we discussed that the number of stories didn't warrant that. 
Closes #91 
cluster


ui/src/components/PinCluster.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from "react";
+import L from "leaflet";
+import resting from "../assets/resting.svg";
+import { Pin, PinState } from "./Pin";
+import "react-leaflet-markercluster/dist/styles.min.css";
+import "leaflet.markercluster";
+import MarkerClusterGroup from "react-leaflet-markercluster";
+import "leaflet/dist/leaflet.css";
+import "react-leaflet-markercluster/dist/styles.min.css";
```
lesliexin: Duplicate import

```suggestion

```

lesliexin (approved):
Tested locally and works well! Just left a small comment

ui/src/components/PinCluster.tsx
```diff
@@ -0,0 +1,40 @@
...
+      showCoverageOnHover={false}
+      spiderLegPolylineOptions={{ opacity: 0 }}
+      iconCreateFunction={createClusterCustomIcon}
+    >
+      {clusterPositions.map((marker, key) => {
+        return (
+          <Pin
+            key={key}
+            position={[marker.lat, marker.lng]}
+            state={PinState.Resting}
```
aspensantos, line 33: Is there a reason for the state of the pin to be a prop if we always use the same initial state?

hanlinc27: I copied the same format as before, plus it wouldn't always be Resting I think there'd also be redirect states in the future?

hanlinc27:
Need to merge #111 to fix lint issues first.

ui/src/components/PinCluster.tsx
aspensantos: ah okay I was just wondering, looks good to me then!

aspensantos (approved):

PR #104, opened 2020-11-12
schema-change

### LINKED ISSUE #84: Migrate CurrentCity field from Authors table to Stories table
- An author could move cities and still publish stories hence we need to have `CurrentCity` as a field in stories

Shehryar21:
closes #84 
- Removed current_city field from Authors table.
- Added current_city and summary field in Stories table.
- Changed unit tests to incorporate new changes
- Changed migrations

### DIFF
--- a/internal/database/migrations/migrations.go
+++ b/internal/database/migrations/migrations.go
@@ -35,7 +35,7 @@ func Seed(db *gorm.DB) error {
 				Title:       faker.Hacker().Noun(),
 				Content:     faker.Hacker().SaySomethingSmart(),
 				CurrentCity: "Toronto",
-				Summary:     "Summary",
+				Summary:     faker.Hacker().SaySomethingSmart(),
 				Latitude:    float64(faker.Address().Latitude()),
 				Longitude:   float64(faker.Address().Longitude()),
 			},
--- a/internal/database/migrations/migrations.go
+++ b/internal/database/migrations/migrations.go
@@ -30,13 +30,14 @@ func Seed(db *gorm.DB) error {
 		LastName:      faker.Name().LastName(),
 		Bio:           faker.Name().Title(),
 		OriginCountry: faker.Address().Country(),
-		CurrentCity:   "Toronto",
-		Latitude:      float64(faker.Address().Latitude()),
-		Longitude:     float64(faker.Address().Longitude()),
 		Stories: []models.Story{
 			{
-				Title:   faker.Hacker().Noun(),
-				Content: faker.Hacker().SaySomethingSmart(),
+				Title:       faker.Hacker().Noun(),
+				Content:     faker.Hacker().SaySomethingSmart(),
+				CurrentCity: "Toronto",
+				Summary:     "Summary",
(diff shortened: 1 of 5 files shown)


restapi/endpoints_test.go
```diff
@@ -119,9 +127,13 @@ func (suite *endpointTestSuite) TestGetAllStories() {
 	response.Object().Value("payload").Array().Element(0).Object().Value("ID").Equal(1)
 	response.Object().Value("payload").Array().Element(0).Object().Value("content").Equal("Children")
 	response.Object().Value("payload").Array().Element(0).Object().Value("title").Equal("The Little Prince")
+	response.Object().Value("payload").Array().Element(0).Object().Value("summary").Equal("Summary")
+	response.Object().Value("payload").Array().Element(0).Object().Value("current_city").Equal("Toronto")
 	response.Object().Value("payload").Array().Element(1).Object().Value("ID").Equal(2)
 	response.Object().Value("payload").Array().Element(1).Object().Value("content").Equal("Fiction")
 	response.Object().Value("payload").Array().Element(1).Object().Value("title").Equal("Hitchhiker's Guide to the Galaxy")
+	response.Object().Value("payload").Array().Element(1).Object().Value("summary").Equal("Summary")
```
hanlinc27: super tiny thing, but it might be best to change the summary text to be slightly different? Even "summary1" and "summary2" would be good.

restapi/stories.go
```diff
@@ -15,6 +15,16 @@ import (
 func (api api) ReturnAllStories(w http.ResponseWriter, r *http.Request) render.Renderer {
 	var stories []models.Story
 
+	for i := 0; i < len(stories); i++ {
```
s278gupt: should this be in `CreateStories` endpoint instead? 

if i'm not wrong, this code is setting the `longitude` and `latitude` fields of the story based on its `CurrentCity` field - if this is true, i think we should be doing this when creating stories, not when returning them

internal/database/migrations/migrations.go
```diff
@@ -30,13 +30,14 @@ func Seed(db *gorm.DB) error {
...
-		Latitude:      float64(faker.Address().Latitude()),
-		Longitude:     float64(faker.Address().Longitude()),
 		Stories: []models.Story{
 			{
-				Title:   faker.Hacker().Noun(),
-				Content: faker.Hacker().SaySomethingSmart(),
+				Title:       faker.Hacker().Noun(),
+				Content:     faker.Hacker().SaySomethingSmart(),
+				CurrentCity: "Toronto",
+				Summary:     "Summary",
```
s278gupt: wdyt about using `faker` to generate random string for this field as well?

restapi/stories.go
Shehryar21: Yes you are right. It should be in createStories. I will fix it.

s278gupt (approved):
lgtm :) 

p.s. feel free to re-request changes (the refresh button next to reviewer name) when you feel like your pr has addressed all comments/is ready for another review

PR #122, opened 2020-11-22
add image_url field

### LINKED ISSUE #117: Add Url field in stories table
URL field will hold the link to image

Shehryar21:
closes #117

### DIFF
--- a/internal/database/migrations/migrations.go
+++ b/internal/database/migrations/migrations.go
@@ -38,6 +38,7 @@ func Seed(db *gorm.DB) error {
 				Summary:     faker.Hacker().SaySomethingSmart(),
 				Latitude:    float64(faker.Address().Latitude()),
 				Longitude:   float64(faker.Address().Longitude()),
+				ImageURL:    "ExampleURL",
 			},
 		},
 	}
--- a/internal/database/migrations/migrations.go
+++ b/internal/database/migrations/migrations.go
@@ -38,7 +38,7 @@ func Seed(db *gorm.DB) error {
 				Summary:     faker.Hacker().SaySomethingSmart(),
 				Latitude:    float64(faker.Address().Latitude()),
 				Longitude:   float64(faker.Address().Longitude()),
-				ImageURL:    "ExampleURL",
+				ImageURL:    "https://exampleurl.com",
 			},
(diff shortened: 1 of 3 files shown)


internal/database/migrations/migrations.go
```diff
@@ -38,6 +38,7 @@ func Seed(db *gorm.DB) error {
 				Summary:     faker.Hacker().SaySomethingSmart(),
 				Latitude:    float64(faker.Address().Latitude()),
 				Longitude:   float64(faker.Address().Longitude()),
+				ImageURL:    "ExampleURL",
```
dhillondeep: make the sample input into a URL format. Something like `https://exampleurl.com`. This is important because we can run validation to make sure URL is valid or not

internal/database/models/models.go
```diff
@@ -25,6 +25,7 @@ type Story struct {
 	Summary         string  `gorm:"type:text" json:"summary"`
 	Latitude        float64 `gorm:"not null" json:"latitude"`
 	Longitude       float64 `gorm:"not null" json:"longitude"`
+	ImageURL        string  `json:"image_url"`
```
dhillondeep: make this `type:text` because URLs can be much longer and text should fit it all. Limit for a normal string is something like 255 char

dhillondeep (changes requested):
about the comments I left

s278gupt:
lgtm, i think we should add this to the happy path test for [creating story](https://github.com/uwblueprint/shoe-project/blob/main/restapi/endpoints_test.go#L236) and [getting story](https://github.com/uwblueprint/shoe-project/blob/main/restapi/endpoints_test.go#L257)

dhillondeep:
> lgtm, i think we should add this to the happy path test for [creating story](https://github.com/uwblueprint/shoe-project/blob/main/restapi/endpoints_test.go#L236) and [getting story](https://github.com/uwblueprint/shoe-project/blob/main/restapi/endpoints_test.go#L257)

Add tests for the model? Yeah, right we can have that as test data for those tests. @Shehryar21 wanna add that real quick and then we can get this merged?

dhillondeep (approved):

Issue #85, opened 2020-11-01
Extend authentication to include roles


dhillondeep:
This is to support multi roles in the authentication system. Description of roles is below

### Owner
- There can be multiple owners
- Owner has all privileges and can add and remove admin users
- Owner can change privileges for admin users

### Admin
- There can be multiple admin users (start with 0 admin users)
- Admin has by default only **add-stories** privilege
- Admin privileges can only be defined by Owner

## Privileges
- **add-stories** - allows a user to add stories using the portal
- **delete-stories** - allows a user to remove stories using the portal
- **edit-preview** - allows a user to change display setting for stories using the portal

PR #134, opened 2020-11-28
Boilerplate user roles using Casbin

s278gupt:
begins work on: https://github.com/uwblueprint/shoe-project/issues/85

the pr configures Casbin and the middleware that checks if a user has the permissions to access a route

_(will soon add a description explaining how the pr uses Casbin to implement user roles & permissions)_


s278gupt:
After talking to client over winter break, spec has changed. No longer implementing user roles.

PR #137, opened 2020-11-29
Pull stories from backend

### LINKED ISSUE #93: Pull stories info and content from the backend
Currently, the data that is being shown on the front-end is static. We need to fetch data from the backend and show it

DinuWije:
## Description
Pulling mass story data from the backend and displaying pins on the map based on the current location (lat/long) of the author.

## To-do:
Update pin preview and story drawer with data from the backend

## Demo
image

### DIFF
--- a/ui/src/pages/ShoeMap.tsx
+++ b/ui/src/pages/ShoeMap.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
 import styled from "styled-components";
+import useSWR from "swr";
 import { AttributionControl, Map, TileLayer, ZoomControl } from "react-leaflet";
 import Control from "react-leaflet-control";
 import "leaflet/dist/leaflet.css";
@@ -9,6 +10,7 @@ import { useState } from "react";
 import { colors } from "../styles";
 import ShoeLogo from "../assets/images/shoeproject-logo.svg";
 import { Filter } from "../components/Filter";
+import { Story } from "../types";
 
 const StyledMap = styled(Map)`
   height: 100vh;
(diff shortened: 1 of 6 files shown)


ui/src/pages/ShoeMap.tsx
```diff
@@ -92,6 +80,15 @@ export const ShoeMap: React.FC = () => {
   const currentLocation = { lat: 43.4723, lng: -80.5449 };
   const [isDrawerOpen, setIsDrawerOpen] = useState(StoryDrawerState.Closed);
 
+  const { data:stories, error } = useSWR<Story[]>("/api/stories");
+
+  var markerList = []
+  if(stories && !error){
+    stories.forEach(story => {
+      markerList.push({lat: story.latitude, lng: story.longitude})
```
AbhiPrasad: Can we just change the expected markerList type to be stories instead of creating a new array.

You'll have to change the types that are passed down into the `PinCluster` component

ui/src/types/index.tsx
```diff
@@ -11,6 +11,8 @@ interface Model {
 export interface Story extends Model {
   title: string;
   content: string;
+  latitude: number;
+  longitude: number;
```
AbhiPrasad, line 17: Can we update this type to have all the story fields from the golang model https://github.com/uwblueprint/shoe-project/blob/main/internal/database/models/models.go#L20?

AbhiPrasad (changes requested):
Rather than create a new list with `var markerList = []`, just update the prop passed down into the `PinCluster` component to accept a list of stories.

AbhiPrasad (approved):

PR #139, opened 2020-12-01
feat: Welcome Screens

### LINKED ISSUE #40: Welcome & Instructions View

### LINKED ISSUE #121: Create Welcome Tutorial components
As a user, I should see the welcome tutorial start when I <br/> 1) First land on the map from the shoe project website's redirect <br/> 2) the **Launch Tour** button on help panel is clicked.

hanlinc27:
### Description
Component for the tutorial frames that happen when launch button is clicked from the help panel and also when a user first lands on the Shoe Map.
[Figma Link](https://www.figma.com/file/ga8lgXwbdArh0Mo1iTIfcx/Designs?node-id=463%3A9091)

### Demo
ezgif com-gif-maker

### Next Steps
* I'm not entirely sure how to get a spotlight going on the pins and filter, right now it looks like this: 

When we want it to look like this: 

Will do some googling but off the top of my head maybe set the MUI dialog's backdrop to transparent and then set some sort of custom overlay..?

* Don't know why the logo on the first screen seems to be a little cut off on the bottom (checked padding and margins already)
* future integration --> Need to make sure that map is centered when welcome tutorial is launched so overlays are in correct positions

Closes  #121


hanlinc27:
Will clean up the repetitive CSS and rebase on main during work session :)

ui/src/pages/ShoeMap.tsx
```diff
@@ -131,6 +132,7 @@ export const ShoeMap: React.FC = () => {
         currentCity={"Toronto"}
         setState={setIsDrawerOpen}
       />
+      <WelcomeTutorial></WelcomeTutorial>
```
AbhiPrasad:
```suggestion
      <WelcomeTutorial/>
```

ui/src/components/WelcomeTutorial.tsx
```diff
@@ -0,0 +1,431 @@
...
+    border-radius: 10px;
+    overflow-y: visible;
+    overflow-x: visible;
+    margin: 0px;
+    top: 20%;
+  }
+
+  .MuiBackdrop-root {
+    opacity: 10%;
+    // background-color: transparent;
```
AbhiPrasad:
```suggestion
``` 

Can we remove?

ui/src/components/WelcomeTutorial.tsx
```diff
@@ -0,0 +1,431 @@
...
+    box-shadow: none;
+    background-color: ${colors.grey};    
+`;
+
+const StyledNavigateContainer = styled.div`
+  display: inline-block;
+`;
+
+export function WelcomeTutorial() {
+  const [open, setOpen] = React.useState("first");
```
AbhiPrasad: Let's move these strings, `first`, `second`, into a string enum.

That way we can refer to it as `Tutorial.First`, `Tutorial.Second` etc.

ui/src/components/WelcomeTutorial.tsx
```diff
@@ -0,0 +1,431 @@
...
+  const handleClose = () => {
+    setOpen(null);
+  };
+
+  return (
+    <>
```
AbhiPrasad: Hmm, can we prefer `React.Fragment` syntax here? Just to make it more clear to people who are new to JSX/frontend.

ui/src/components/WelcomeTutorial.tsx
```diff
@@ -0,0 +1,431 @@
...
+  NavigateTitleText,
+  NavigateDescriptionText,
+} from "../styles/typography";
+
+const StyledArrowTip = styled.img`
+  position: absolute;
+  width: 29px;
+  height: 26px;
+  left: -5%;
+  top: 45%;
```
AbhiPrasad, line 25: Do we have no other way other than `position: absolute` + manually ajdusting element with css? Maybe with something like https://github.com/shipshapecode/react-shepherd?

ui/src/components/WelcomeTutorial.tsx
AbhiPrasad: Hmm, also thinking about this more, should the tutorial be handling it's own state? Maybe we should lift it up to the parent component

AbhiPrasad (changes requested):
We'll want to only run this welcome screen once, maybe we can store a localstorage cookie? (timed like a month so that it'll refresh after a while). We can move that into it's own ticket tho.

Nice implementation, left some thoughts on some areas though.

AbhiPrasad:
In terms of the spotlight on the filter + pin, we can grab a dom reference through those elements and store it in the page component (https://reactjs.org/docs/refs-and-the-dom.html). We can then pass that reference down to the `WelcomeTutorial` component and we can render elements accordingly (it'll require some css hacks but shouldn't be that hard).

ui/src/components/WelcomeTutorial.tsx
hanlinc27: The component should be able to handle its own state to go from screen to screen right? I added it as a prop now too so that it can be set from the parent component as well (to launch or close) the tutorial

ui/src/components/WelcomeTutorial.tsx
hanlinc27: That looks pretty cool I wish I knew about it earlier! I know the manual adjustment is a bit suspect (esp the negative overflow) but it's relative + bounded to the dialog so it wouldn't cause issues with responsiveness and it seems to be the simplest solution? Could look at using/converting to that tour library in the future if we end up with a heavier tutorial ^^

hanlinc27: apparently this is a thing too https://docs.react-joyride.com/

ui/src/components/WelcomeTutorial.tsx
AbhiPrasad: Yeah you're right, I think this works for now.

ui/src/components/WelcomeTutorial.tsx
AbhiPrasad: Good point, I guess it is only a 3 step tutorial. Let's use your approach for now then, and we can always change it later!

AbhiPrasad (approved):
Ok sweet, this LGTM.

Opened https://github.com/uwblueprint/shoe-project/issues/147 about the localstorage stuff.

PR #145, opened 2020-12-03
Map arrows

### LINKED ISSUE #90: Create Edge Markers to point towards nearest story
A user should see arrows when no markers are in sight

aspensantos:
Capture

Closes #90


AbhiPrasad:
We worked on https://github.com/uwblueprint/shoe-project/pull/143 to simplify frontend state a little bit.

From skimming over the PR, it doesn't seem too bad to rebase this PR on top of that, but curious to hear your thoughts. If it's a lot of trouble, we can merge this first, and then do the state stuff.

aspensantos:
> We worked on #143 to simplify frontend state a little bit.
> 
> From skimming over the PR, it doesn't seem too bad to rebase this PR on top of that, but curious to hear your thoughts. If it's a lot of trouble, we can merge this first, and then do the state stuff.

Since this only uses the story state it should be okay to rebase, I will do that once the other PR is merged.

ui/src/components/Arrow.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const x = unit * Math.sin(rad);
+  const y = unit * Math.cos(rad);
+
+  const icon =
+    `
+    <svg width='70' height='70' viewBox='-10 -10 80 80' fill='none' xmlns='http://www.w3.org/2000/svg' >
+        <path transform='scale(0.3) translate(` +
+    y +
+    `, ` +
+    x +
```
AbhiPrasad: Can we do string interpolation with template literals here? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

so:

```
<svg width='70' height='70' viewBox='-10 -10 80 80' fill='none' xmlns='http://www.w3.org/2000/svg' >
        <path transform='scale(0.3) translate(`${y}, ${x}) rotate(${rotation} blah blah
```

ui/src/components/EdgePinCluster.tsx
```diff
@@ -0,0 +1,127 @@
...
+import L, { LatLng } from "leaflet";
+import "react-leaflet-markercluster/dist/styles.min.css";
+import "leaflet.markercluster";
+import MarkerClusterGroup from "react-leaflet-markercluster";
+import "leaflet/dist/leaflet.css";
+import { EdgePin, getAnchor } from "./EdgePin";
+import { EdgeType } from "./EdgePins";
+import { arrow } from "./Arrow";
+
+const createClusterCustomIcon = function (edgeType: EdgeType) {
```
AbhiPrasad: Prefer arrow function in this case.

```suggestion
const createClusterCustomIcon = (edgeType: EdgeType) => {
```

ui/src/components/Arrow.tsx
```diff
@@ -0,0 +1,41 @@
+export function arrow(angle: number): string {
```
AbhiPrasad: Can we make this func name a little more descriptive?

ui/src/components/EdgePin.tsx
```diff
@@ -0,0 +1,47 @@
...
+    shadowSize: null,
+    shadowAnchor: null,
+    iconSize: new L.Point(60, 70),
+  });
+
+  return (
+    <React.Fragment>
+      <Marker position={position} icon={arrowIcon} />
+    </React.Fragment>
+  );
```
AbhiPrasad:
```suggestion
  return (
    <Marker position={position} icon={arrowIcon} />
  );
```

ui/src/components/EdgePin.tsx
```diff
@@ -0,0 +1,47 @@
...
+export function getAnchor(edgeType: EdgeType): Point {
+  switch (edgeType) {
+    case EdgeType.Top:
+      return new L.Point(0, -10);
+    case EdgeType.Bottom:
+      return new L.Point(0, 110);
+    case EdgeType.Left:
+      return new L.Point(-30, 0);
+    default:
+      return new L.Point(80, 0);
```
AbhiPrasad, line 25: Where are these numbers coming from? If they're just based on experiments, can we leave a comment explaining how we got there?

ui/src/components/EdgePinCluster.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const bottomCluster = clusterData.filter(
+    (data) => data.edgeType === EdgeType.Bottom
+  );
+  return (
+    <React.Fragment>
+      <MarkerClusterGroup
+        showCoverageOnHover={false}
+        spiderLegPolylineOptions={{ opacity: 0 }}
+        iconCreateFunction={() => createClusterCustomIcon(EdgeType.Right)}
+      >
```
AbhiPrasad: Hmm can we just do this in a loop if everything stays the same but `EdgeType` and the array thats being mapped over?

Something like (sorry for bad formatting lol):

```ts
const arr = [{type: EdgeType.Right, cluster: rightCluster}, {type: EdgeType.Left, cluster: leftCluster} ... etc]
arr.map(({ type, cluster }) => {
  <MarkerClusterGroup
        showCoverageOnHover={false}
        spiderLegPolylineOptions={{ opacity: 0 }}
        iconCreateFunction={() => createClusterCustomIcon(type)}
      >
  {cluster.map((marker, key) => {
          return (
            <EdgePin
              key={key}
              position={marker.pos}
              angle={marker.angle}
              edgeType={marker.edgeType}
            />
          );
        })}
   </MarkerClusterGroup>
});
```

ui/src/components/EdgePins.tsx
```diff
@@ -0,0 +1,159 @@
...
+}
+
+function intersects(a, b, c, d, p, q, r, s) {
+  const det = (c - a) * (s - q) - (r - p) * (d - b);
+  if (det === 0) {
+    return false;
+  } else {
+    const lambda = ((s - q) * (r - a) + (p - r) * (s - b)) / det;
+    const gamma = ((b - d) * (r - a) + (c - a) * (s - b)) / det;
+    return 0 < lambda && lambda < 1 && 0 < gamma && gamma < 1;
```
AbhiPrasad, line 30: Mind writing a comment how this works?

ui/src/components/EdgePins.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  const pinPositions = stories.map((story) => {
+    return { lat: story.latitude, lng: story.longitude };
+  });
+
+  pinPositions.forEach((pos) => {
+    if (mapBounds.contains(L.latLng(pos.lat, pos.lng))) {
+      isVisible = false;
+    }
+  });
```
AbhiPrasad: Lets use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some here.

```suggestion
const isVisible = pinPositions.some((pos) => mapBounds.contains(L.latLng(pos.lat, pos.lng)));
```

ui/src/components/EdgePins.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  // do not display the edge pins if there are markers in view
+  if (!isVisible) return null;
+
+  // determine the distance from the center to each of the markers
+  const distanceToPos: { [distance: number]: LatLng } = {};
+  pinPositions.forEach((pos) => {
+    const posLatLng = L.latLng(pos.lat, pos.lng);
+    distanceToPos[currPosition.distanceTo(posLatLng)] = posLatLng;
```
AbhiPrasad, line 136: Nit: we can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce to create an object. Ping me if you got questions about it.

Also fine with the current approach.

ui/src/components/EdgePins.tsx
```diff
@@ -0,0 +1,159 @@
...
+  });
+
+  //sort the distances and take the first 3
+  const distances = Object.keys(distanceToPos);
+  distances.sort().slice(0, 3);
```
AbhiPrasad: Slice doesn't mutate the array, does this do what you want it to do?

ui/src/components/EdgePins.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  // get the closest three pins
+  const closestPins = [];
+  distances.forEach((distance) => {
+    closestPins.push(distanceToPos[distance]);
```
AbhiPrasad: We can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map here

ui/src/components/EdgePins.tsx
```diff
@@ -0,0 +1,159 @@
...
+  // calculate intersection of the line from the center to the pin
+  // store that point of intersection and the angle of the direction vector in edgePinsInfo
+  const corners = [
+    mapBounds.getSouthWest(),
+    mapBounds.getSouthEast(),
+    mapBounds.getNorthEast(),
+    mapBounds.getNorthWest(),
+  ];
+
+  closestPins.forEach((pin) => {
```
AbhiPrasad, line 158: Can we pull this out into it's own function? Maybe even add some comments on what it is doing?

ui/src/components/EdgePins.tsx
```diff
@@ -0,0 +1,159 @@
...
+export enum EdgeType {
+  Bottom,
+  Right,
+  Top,
+  Left,
+}
+
+function getEdgeType(index) {
+  switch (index) {
+    case 0:
```
AbhiPrasad: We don't need this, we can just init the enum type and use it.

```ts
export enum EdgeType {
  Bottom = 0,
  Right,
  Top,
  Left,
}
```

then 
```ts
EdgeType.Bottom === 0 // true
```

AbhiPrasad (changes requested):
Actual math wow. Nice work figuring this out!

Left some comments, mostly about javascript things. Ping me if you got questions about specifics.

hanlinc27 (commented):
Can you update the redirect icon and styling before merge as well? 😃 @m-penny

ui/src/components/EdgePins.tsx
```diff
@@ -0,0 +1,154 @@
...
+  pinPositions.forEach((pos) => {
+    const posLatLng = L.latLng(pos.lat, pos.lng);
+    distanceToPos[currPosition.distanceTo(posLatLng)] = posLatLng;
+  });
+
+  //sort the distances and take the first one
+  const distances = Object.keys(distanceToPos).sort().slice(0, 1);
+
+  // get the closest three pins
+  const closestPins: LatLng[] = distances.map((distance) => {
```
AbhiPrasad, line 143: Isn't distances is a single element array because of `slice(0, 1)`?

ui/src/components/Arrow.tsx
AbhiPrasad: This still needed?

AbhiPrasad (approved):
I think we good to merge after CI is green (left 2 comments)

ui/src/components/Arrow.tsx
aspensantos: Yes but I updated it with the new styling

aspensantos:
> Can you update the redirect icon and styling before merge as well? 😃 @m-penny

Yes, updated :)

ui/src/components/EdgePins.tsx
aspensantos: Just left it got tired 🥴

ui/src/components/EdgePins.tsx
aspensantos: Yeah this is because design changed it from the 3 closest to the closest one, so I made it so we could have more than one, but now it is just one. Not sure if I should leave it in case we want more in the future?

dhillondeep:
These changes caused few bugs and slowdown on the map so, I rollbacked the changes and we can deploy this again when these issues are resolved

PR #149, opened 2020-12-06
Filter story pins from user input

DinuWije:
## Description 
Taking user input to filter pins onto the map.

## Demo
image
image

### DIFF
--- a/ui/src/components/Filter.tsx
+++ b/ui/src/components/Filter.tsx
@@ -17,8 +17,7 @@ const icon = <CheckBoxOutlineBlankIcon fontSize="small" />;
 const checkedIcon = <CheckBoxIcon fontSize="small" />;
 
 export interface FilterProps {
-  options: string[];
-  onChange: (options: string[]) => void;
+  onChange: (event: React.ChangeEvent, options: string[]) => void;
 }
 
 const theme = createMuiTheme({
@@ -55,18 +54,12 @@ const Tagline = styled.span`
   padding-bottom: 16px;
 `;
 
-export function Filter({state, setState}): JSX.Element {
+export function Filter({ onChange }: FilterProps): JSX.Element {
   const { data: countries, error } = useSWR<string[]>(
     "/api/authors/origin_countries"
   );
 
   if (error) return <div>Error!</div>;
-
-  function onTagsChange(event, values) {
-    setState(
-      values
-    );
-  }
 
   return (
     <ThemeProvider theme={theme}>
@@ -79,7 +72,7 @@ export function Filter({state, setState}): JSX.Element {
           id="filter-autocomplete"
           options={countries || []}
           style={{ width: 312 }}
-          onChange={onTagsChange}
+          onChange={onChange}
           disableCloseOnSelect
           renderTags={(value, getTagProps) =>
             value.map((option, index) => (
--- a/ui/src/components/Filter.tsx
+++ b/ui/src/components/Filter.tsx
@@ -11,7 +11,6 @@ import CheckBoxOutlineBlankIcon from "@material-ui/icons/CheckBoxOutlineBlank";
 import CheckBoxIcon from "@material-ui/icons/CheckBox";
 import { createMuiTheme } from "@material-ui/core/styles";
 import { ThemeProvider } from "@material-ui/styles";
-
 import { FilterChip } from "./FilterChip";
 
 const icon = <CheckBoxOutlineBlankIcon fontSize="small" />;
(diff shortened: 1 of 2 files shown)


ui/src/components/Filter.tsx
```diff
@@ -56,12 +55,18 @@ const Tagline = styled.span`
   padding-bottom: 16px;
 `;
 
-export function Filter(): JSX.Element {
+export function Filter({state, setState}): JSX.Element {
```
AbhiPrasad: Why do we need to pass `state` into the filter?

Can we just expose `onChange`, and move the `onTagsChange` function up to the parent component? Also please type the props?

so something like:

```
export function Filter({ onChange }: FilterProps): JSX.Element {
```

ui/src/pages/ShoeMap.tsx
```diff
@@ -80,7 +80,7 @@ export const ShoeMap: React.FC = () => {
   const currentLocation = { lat: 43.4723, lng: -80.5449 };
 
   const { data: stories, error } = useSWR<Story[]>("/api/stories");
-
+  const [filteredStories, setFilteredStories] = useState([]);
```
AbhiPrasad: This is not an array of stories, it's an array of countries, so we should probably rename accordingly.

ui/src/pages/ShoeMap.tsx
```diff
@@ -101,9 +101,31 @@ export const ShoeMap: React.FC = () => {
           <TileLayer
             url={`https://api.mapbox.com/styles/v1/hanlinc27/ckhjy5wat2dvz1aplv4tkaghb/tiles/{z}/{x}/{y}?access_token=${process.env.REACT_APP_MAPBOX_ACCESS_TOKEN}`}
           />
-          {stories && !error && (
-            <PinCluster stories={stories} openDrawer={handleOpenDrawer} />
-          )}
+          {(() => {
+
+            if(stories && filteredStories.length != 0){
+              var filteredStoryArr= new Array<Story>();
```
AbhiPrasad: 1) You can find the intersection between two arrays with `.filter` and `.includes`. So something like:
```
const filteredStoryArr = stories.filter(value => filterStories.includes(value.author_country));
```
^ This is not super optimum, but works well for our case. If you are wondering about performance, we can also use es6 sets (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) or memoize with a hash map.

2) We shouldn't need to call a anon function with `(() => {...})()`. Instead we can compute this before the return value. So something like:

```js
const { data, error } = useSWR<Story[]>("/api/stories");
const [filteredCountries, setFilteredCountries] = useState([]);

...

// If no filtered countries just make stories equal to data
// otherwise filter the stories based on the intersection of the two arrays.
const stories = filteredCountries.length === 0 ? data : data.filter(story => filterStories.includes(story.author_country));

...

return (
  ...
  ...
  
  <PinCluster stories={stories} openDrawer={handleOpenDrawer} />
  ...
  ...
```

AbhiPrasad (changes requested):
Nice! Lmk if you have any questions about my review.

Oh yeah make sure you run `yarn fix` in the ui folder to make CI go green (don't merge on red CI).

ui/src/pages/ShoeMap.tsx
```diff
@@ -101,33 +111,10 @@ export const ShoeMap: React.FC = () => {
...
-            
-            else if(stories){
-              return(
-                <PinCluster stories={stories} openDrawer={handleOpenDrawer} />
-              )
-            }
-
-          })()}
+          {stories && !error && (
+            <PinCluster stories={stories} openDrawer={handleOpenDrawer} />
```
AbhiPrasad, line 115: Can we create a Github issue so we keep track of render stuff in error states?

Will require some design work as well so it we don't gotta do it immediately.

ui/src/components/Filter.tsx
```diff
@@ -79,7 +72,7 @@ export function Filter({state, setState}): JSX.Element {
           id="filter-autocomplete"
           options={countries || []}
           style={{ width: 312 }}
-          onChange={onTagsChange}
+          onChange={(event, value) => onChange(value)}
```
AbhiPrasad: Let's change the type here. This means that:

```suggestion
          onChange={onChange}
```

```ts
export interface FilterProps {
  onChange: (event: idkWhatEventTypeIs, value: string[]) => void
}
```

Then you'll have to change:

```ts
// ui/src/pages/ShoeMap.tsx 
function onTagsChange(_, value: string[]) {
  setFilteredCountries(options);
}
```

AbhiPrasad (commented):
1 more thing, I'll approve after this. Good stuff!

ui/src/pages/ShoeMap.tsx
DinuWije: Done!

AbhiPrasad (approved):
🚢 🇮🇹   (make sure to squash and merge)

PR #150, opened 2020-12-06
build: Add eslint rules for hooks and import/export

AbhiPrasad:
Add two eslint plugins:

1) https://www.npmjs.com/package/eslint-plugin-react-hooks to make sure that react hooks are being used correctly

2) https://www.npmjs.com/package/eslint-plugin-simple-import-sort to organize our imports and export properly. You should be able to just run `yarn fix` to organize em, so don't worry about manually trying to organize it.


ui/.eslintrc.js
```diff
@@ -27,5 +28,9 @@ module.exports = {
         "ts-ignore": "allow-with-description",
       },
     ],
```
hanlinc27, line 30: Is it worth adding the additional config for the hooks plugin?
`   "react-hooks/rules-of-hooks": "error", // Checks rules of Hooks `
`   "react-hooks/exhaustive-deps": "warn" // Checks effect dependencies`

hanlinc27 (approved):
The simple-import-sort will be very nice 😄

ui/.eslintrc.js
AbhiPrasad: By extending the recommended eslint config, those rules (and more) should be enabled by default. We can update em later if we find some annoying.

PR #158, opened 2020-12-09
fix: Pin, Pin Preview, and Filter Component Stylings

### LINKED ISSUE #124: Fix/update peer dependencies for UI
Need to rollback the change in #112

### LINKED ISSUE #152: Limit map zoom to Canada
Since there are three main clusters of pins, and per the product spec doc, the leaflet map should be bound to Canada.
MAX: (ZoomMin)
image

### LINKED ISSUE #153: Update Pin Styling
The Pin styling needs to be updated to include the shoe icon + change in size. 
[Figma Link](https://www.figma.com/file/ga8lgXwbdArh0Mo1iTIfcx/Designs?node-id=308%3A1450)
image

hanlinc27:
Just some small changes:
* Updated the pin styling and size for resting, unfocused, hover etc.
* Updated font styling for pin preview
* Fixed some minor styling on Filter Component (changed to styled components to be consistent) Popper text styling, border underline colors, input text styling, cancel button colour
* Changed the zoom so that it lands on middle of Canada when open
* Tried to upgrade some of the dependency warnings

Resolves #153, Resolves #152, Resolves #124 

image
image


ui/src/components/Filter.tsx
```diff
@@ -54,6 +83,17 @@ const Tagline = styled.span`
   padding-bottom: 16px;
 `;
 
+const useStyles = makeStyles(() => ({
+  paper: {
```
hanlinc27, line 84: This was the only way I was able to get the Popper text styled correctly even though I know we want to mainly use styled components
image
Wasn't able to override the css properly with the ` .MuiAutcomplete-paper` class

ui/src/components/PinCluster.tsx
```diff
@@ -38,7 +38,7 @@ export const PinCluster = React.memo(function PinCluster({
             id={story.ID.toString()}
             key={story.ID}
             story={story}
-            state={PinState.Resting}
+            state={PinState.Unfocused}
             onPopupClick={openDrawer(story)}
```
hanlinc27, line 42: Will probably need to open up an issue about the pin hover behavior as well

AbhiPrasad (approved):
🚢 🇮🇹

PR #165, opened 2020-12-13
Fetch client facing tokens from backend

dhillondeep:
- [x] Fetch client facing tokens from backend
- [x] Use fetched tokens in UI


restapi/api.go
```diff
@@ -37,6 +37,8 @@ func Router(db *gorm.DB, locationFinder location.LocationFinder) (http.Handler,
 		rest.GetHandler(r, "/story/{storyID}", api.ReturnStoryByID)
 		rest.GetHandler(r, "/authors/origin_countries", api.ReturnAllCountries)
 		rest.PostHandler(r, "/login", api.Login)
+
+		rest.GetHandler(r, "/client-tokens", api.ReturnClientTokens)
```
AbhiPrasad: We use `/authors/origin_countries` but here it is `/client-tokens`. We should standardize between `_` and `-`

AbhiPrasad (approved):

Shehryar21 (approved):

PR #168, opened 2020-12-13
Heroku deployment

Shehryar21:


config/getter.go
```diff
@@ -81,6 +81,15 @@ func GetProfilerPath() string {
 
 // -- server --
 
+func GetPort() *int {
+	if !viper.IsSet("port") {
```
AbhiPrasad, line 85: Can’t we set a default port in code and return an int instead of a pointer?

dhillondeep (approved):

AbhiPrasad (commented):
Golang needs actual tagged union types ngl.

Imagine golang with [Rust Optionals](https://doc.rust-lang.org/rust-by-example/std/option.html) or [Haskell Maybe](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Maybe.html)

config/getter.go
dhillondeep: yeah but it will be hard for us to tell if $PORT env is provided to us

dhillondeep:
@AbhiPrasad lol sorry I merged the code didn't know you were still reviewing

config/getter.go
AbhiPrasad: Ah word, that’s a good point.

PR #172, opened 2020-12-17
randomize location, fix date styling, fix drawer content styling

### LINKED ISSUE #151: bug: Zoom in on pin cluster when clicked
Right now it looks like (Interaction 1/2)
ezgif com-gif-maker (1)

But we would like it to look like the last pin cluster click.
image (1)

hanlinc27:
@dhillondeep should fix the spiral issue?
### Description
Thoughts on randomizing the coordinates slightly on the client side so that the clusters can be zoomed in?
Closes #151 
cluster

Preview date styling: 
image

Add in the new lines on story drawer content: 
image

### DIFF
--- a/ui/src/components/Pin.tsx
+++ b/ui/src/components/Pin.tsx
@@ -28,6 +28,12 @@ const StyledPopup = styled(Popup)`
   }
 `;
 
+function randomCoords(coordinate: number): number{
+  var num = Math.floor(Math.random()*99) + 1;
+  num *= Math.round(Math.random()) ? 1 : -1; 
+  return (coordinate+0.1*num);
+}
+
 export enum PinState {
   Resting,
   Selected,
@@ -59,7 +65,7 @@ export function Pin({
   });
 
   return (
-    <Marker position={[story.latitude, story.longitude]} icon={icon}>
+    <Marker position={[randomCoords(story.latitude), randomCoords(story.longitude)]} icon={icon}>
       <StyledPopup>
         <PinPreview
           shoeImage={story.image_url}
--- a/ui/src/components/Pin.tsx
+++ b/ui/src/components/Pin.tsx
@@ -28,10 +28,10 @@ const StyledPopup = styled(Popup)`
   }
 `;
 
-function randomCoords(coordinate: number): number{
-  var num = Math.floor(Math.random()*99) + 1;
-  num *= Math.round(Math.random()) ? 1 : -1; 
-  return (coordinate+0.01*num);
+function randomCoords(coordinate: number): number {
+  let num = Math.floor(Math.random() * 99) + 1;
+  num *= Math.round(Math.random()) ? 1 : -1;
+  return coordinate + 0.01 * num;
 }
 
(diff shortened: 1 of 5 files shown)


ui/src/components/PinPreview.tsx
```diff
@@ -72,7 +72,7 @@ export function PinPreview({
           Read Full Story
         </StyledButton>
         <CardDetailText>
-          {author} • {date}
+          {author} • {date.substring(0,4)}
```
hanlinc27: So that only shows the year - but I think we're changing this in the DB anyways?

dhillondeep: yeah we will just have a year as a field in database

dhillondeep (approved):
This looks awesome! We should get this merged. It also fixes text formatting issues soo that is nice

ui/src/components/Pin.tsx
```diff
@@ -59,7 +65,10 @@ export function Pin({
   });
 
   return (
-    <Marker position={[story.latitude, story.longitude]} icon={icon}>
+    <Marker
+      position={[randomCoords(story.latitude), randomCoords(story.longitude)]}
```
AbhiPrasad, line 69: This doesn't make them unique though, there can still be instances where the positions are the same. Should we guard against those?

hanlinc27: I kind of don't like this solution at all to be honest since it makes the viewing experience inconsistent, it's just sort of a hacky way to get the clustering working properly. I think in the future we should either
1) let the user place the pin where they would like in a specific city
2) have them enter a general city location, geocode, and then pick somewhere [random ](https://gis.stackexchange.com/questions/25877/generating-random-locations-nearby)within a specific radius? 

We could also make it ordered and just have it fan out directionally from the original coordinate but I don't think that would look as nice. For now, I think since it's a temporary fix, the chances of it not being unique is very, very small?

dhillondeep: I agree with hanlin on this. We need a fresh design of how coordinates are handled and shown. Current way could work for few stories but when number of stories increase, we could see major trouble. Since the term is finishing I guess this fix could work and next term we could add a proper solution

AbhiPrasad (approved):

ui/src/components/PinPreview.tsx
```diff
@@ -72,7 +72,7 @@ export function PinPreview({
           Read Full Story
         </StyledButton>
         <CardDetailText>
-          {author} • {date}
+          {author} • {date.substring(0, 4)}
```
lesliexin: [nit] if we're getting the `createdAt` value as a typescript `Date` object can we use `dateObj.getYear()` instead?

lesliexin (approved):
🚀

ui/src/components/PinPreview.tsx
hanlinc27: I think I'm just going to revert it entirely since Deep is changing it in the database to just year so it can just fetch it as a string

PR #195, opened 2021-01-23
Fix Seeding

nimwijetunga:


restapi/stories.go
```diff
@@ -86,7 +86,7 @@ func (api api) ReturnStoriesByCountries(w http.ResponseWriter, r *http.Request)
 	return rest.JSONStatusOK(stories)
 }
 
-func randomCoords(coordinate float64) float64 {
+func RandomCoords(coordinate float64) float64 {
```
AbhiPrasad, line 89: What do you think about moving this out into a helper instead of exposing a public member from the current file?

nimwijetunga:
Closing as we no longer need to do this after pin placement is fixed

Issue #185, opened 2021-01-17
Multimedia


nimwijetunga:
- Videos are on Youtube (backend we need a new column in DB for yt link)
  - embed YT on the frontend

- Investigate where to host audio for free (most likely google drive) and get public link
      - Whoever is working on this needs to check if google drive works to host and stream audio, if not whats the alternative
      - embed audio file on the frontend

- Decide whether to continue hosting images on S3 or do it on something like G-Drive (theyll have to host it)
   - Be able to upload an image to S3 from the backend when the image is given by the frontend
   - If it's too much work to upload images from the backend then consider using G-Drive (research)

Shehryar21:
Update: Audio files on GDrive can be used in the frontend

s278gupt:
some questions after doing a lil digging online:

1. Do the multimedia (video, audio, image) columns belong to the `Story` object? And I'm assuming it's one column per link? (I can do a quick pr once this is answered)

>  Be able to upload an image to S3 from the backend when the image is given by the frontend
2. Will this be part of the functionality of an existing post handler, or are we creating a new endpoint specifically for uploading images?
3. There's a couple ways we could go about this. What are your thoughts:
_way 1_:
- frontend uploads image to our server? (where do we store file when we pass it to backend? excuse my ignorance, haven't worked with passing files through api before)
- backend handles the uploading to s3

_way 2_:
- frontend sends `filename` + other fields in request to backend
- backend generates presigned url and sends it as response to frontend
- frontend uses response to request upload to s3

_way 3_:
- entirely frontend 
(i know this is a terrible option, secrets and all, just wanted to be exhaustive in all the options we have)

nimwijetunga:
One thing to note is that we will no longer be doing audio after discussing with the client

1. Yes images and video belong to the story object so we need to add a new column to the model to store a youtube link (video). However we also now have that one story can have multiple images so we need to migrate the database so that the image_url column is now an array (instead of a single value) 

2. No this will be part of the existing story upload endpoint (the user can enter story details as well as pass images)

3. @Shehryar21 is currently looking into doing it using way 1 (essentially we store the image in memory on the backend once we receive it). Ideally we don't want to do way 3 because of security reasons (may expose the s3 bucket). @s278gupt can you look into way 2 and see how easy/hard it is to implement. We will decide (depending on engineering effort) to either go with way 1 or 2 -> One thing to consider for way 2 is how to track which images the user uploaded for a given story (so we can add it to the database)

PR #203, opened 2021-01-27
Adds video link field to story model

s278gupt:
Part 1 of https://github.com/uwblueprint/shoe-project/issues/185

See https://github.com/uwblueprint/shoe-project/issues/185#issuecomment-765444247: `we need to add a new column to the model to store a youtube link (video)`


AbhiPrasad (commented):
Mind also updating the TypeScript types found at https://github.com/uwblueprint/shoe-project/blob/main/ui/src/types/index.tsx#L11?

nimwijetunga:
Also please add some random youtube links to the stories.json file so we can test it once the frontend is ready

s278gupt:
@nimwijetunga feel free to merge this if approved

hanlinc27 (commented):
In the future, we should remember to be adding the embedded youtube URL (found when you click share video) and is the form "youtube.com/embed/...." rather than the watch URL

data/stories.json
```diff
@@ -1,431 +1,437 @@
...
+        "author": {
+            "first_name": "Tanaz",
+            "last_name": "Bhathena",
+            "origin_country": "India"
+        },
+        "video_url": "https://www.youtube.com/watch?v=ao6jHx27gB8",
```
nimwijetunga: can you change this to the embed link (take the parameter v and place it in the url path)
i.e https://www.youtube.com/watch?v=ao6jHx27gB8 -> https://www.youtube.com/embed/ao6jHx27gB8

nimwijetunga: Also please do this for the other youtube links

nimwijetunga (changes requested):

data/stories.json
AbhiPrasad: Hmm should we validate for this? Like say the video URL must follow a certain regex or the POST is not accepted?

nimwijetunga: We cant expect shoe project to do this b/c its kind of annoying to get this url. Instead given a regular youtube url we can convert it to the embed version by extracting the query parameter value for `v`. I think that should be a task for the story upload form (on the backend)

nimwijetunga (approved):

Issue #205, opened 2021-01-30
Login - Frontend


nimwijetunga:
Page 1 - Log in with google
   - link this to our app in google console
   - call /login endpoint to get url to use for the sign in with google button
   - https://www.figma.com/file/uVqnweL2ApqvPHWzQYmZ1t/Dashboard?node-id=66%3A738

Page 2 - Error Page 
  - for invalid emails (not shoeproject or blueprint emails)
  - https://www.figma.com/file/uVqnweL2ApqvPHWzQYmZ1t/Dashboard?node-id=126%3A1780

Wait: We will need to store the JWT token we get from the backend (either session storage or as a cookie)
  - will be done on the all stories page

PR #210, opened 2021-02-01
feat: Auth Provider setup in frontend

AbhiPrasad:
Addresses: https://github.com/uwblueprint/shoe-project/issues/205

## About
This is a small PR for setting up a boilerplate auth setup in the frontend. Figured I could bang this out cause it can be a pain to setup.

Right now we haven't setup OAuth, so the signin and signout functions do nothing.

We leverage the React Context API alongside some custom hooks to make sure we can protect routes under Auth. A lot of the logic was based on the template from https://usehooks.com/useAuth/, but I changed some stuff up for our specific setup.

## Testing
Can't test because the routes (`/login` and `/admin`) render null for now. Once we implement OAuth for real and add actual signin and signout functions we can test.

## Next Steps
1. OAuth proper
2. Write login page
3. Write admin page


ui/src/__tests__/util/timestamp.test.ts
```diff
@@ -2,7 +2,7 @@ import {
   isTimestampExpired,
   setTimeStamp,
   tutorialKey,
-} from "../components/helpers/welcomeTutorialFunctions";
+} from "../../util/timestamp";
```
AbhiPrasad, line 5: I reorganized folder structure so things make a little more sense

ui/src/types/index.tsx
```diff
@@ -35,3 +35,7 @@ export interface Author extends Model {
 export interface Tokens {
   mapbox: string;
 }
+
+export interface User extends Partial<Model> {
+  username: string;
```
AbhiPrasad, line 40: This will change, just put some temporary type for now

DinuWije (approved):

PR #211, opened 2021-02-03
feat: Use Clear button instead of X for filter

### LINKED ISSUE #101: Replace X button with clear button on filter component
* Replace X Button to Clear button
* Clear button need to shift down even when 5+ stories are loaded

AbhiPrasad:
Closes #101 

image

Basically because the Filter is a controlled component, we have to hide the actual X button (with display: none), and then make our CLEAR button click that one. It's not ideal, but it works :))))

### DIFF
--- a/ui/.eslintrc.js
+++ b/ui/.eslintrc.js
@@ -29,6 +29,8 @@ module.exports = {
         "ts-ignore": "allow-with-description",
       },
     ],
+    // Allow for variables prepended with _ to be ignored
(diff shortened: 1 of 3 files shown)


AbhiPrasad:
@DinuWije im sorry if this causes a bad rebase with the mobile stuff

DinuWije (approved):

hanlinc27 (approved):

PR #214, opened 2021-02-06
Upload stories endpoint

### LINKED ISSUE #206: Story Upload - Backend
- add any fields that are currently missing (yt link, etc..)

- define the API for story upload  POST request
  - { 
       'video_url': yt_link,
       ....
     }

- adding logic to upload image that was given from frontend to S3 
   - look at @Shehryar21  pr to see how this is done
  
- create author in the author table if one does not already exist for the story

- To add tags to stories
   - how to store these in the backend (an array of tags per story)
  - keep a separate table in the backend for tags (one column is story id and second column is the tag -> primary key (story_id, tag))

- Convert regular youtube link to embed url (i.e https://www.youtube.com/watch?v=ao6jHx27gB8 -> https://www.youtube.com/embed/ao6jHx27gB8

Post Request Body Format
------------------------------------

Sending data as Form-Data
image: [file],
video_url: "",
title: "",
content: "",
summary: "",
author_first_name: "",
author_last_name: "",
author_country: "",
year: "",
current_city: "",
bio: "",
tags, []

Each tag would be added to another table with the story id and tag name.

Shehryar21:
Created a new endpoint (story). I made a new one instead of changing the existing stories endpoint. This adds a new story and also uploads image to s3. If author is not present, it creates the new author as well.
Partially closes #206


restapi/stories.go
```diff
@@ -72,6 +80,106 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
 	return rest.MsgStatusOK("Stories added successfully")
 }
 
+func (api api) CreateStoriesFormData(w http.ResponseWriter, r *http.Request) render.Renderer {
+
+	awsAccessKeyID := os.Getenv("B2_KEY_ID")
+	awsSecretAccessKey := os.Getenv("B2_APP_KEY")
+	s3Config := &aws.Config{
```
AbhiPrasad: lets move all constants outside of the endpoint. We can add them to the api interface in case other route handlers want access to them.

Constants: `awsAccessKeyID`, `awsSecretAccessKey`, `s3Config`

restapi/stories.go
```diff
@@ -72,6 +80,106 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+		Credentials:      credentials.NewStaticCredentials(awsAccessKeyID, awsSecretAccessKey, ""),
+		Endpoint:         aws.String(os.Getenv("BUCKET_ENDPOINT")),
+		Region:           aws.String(os.Getenv("BUCKET_REGION")),
+		S3ForcePathStyle: aws.Bool(true),
+	}
+	newSession, err := session.NewSession(s3Config)
+	if err != nil {
+		return rest.ErrInvalidRequest(api.logger, "Error connecting with AWS S3 Bucket", err)
+	}
+	s3Client := s3.New(newSession)
```
AbhiPrasad: Can we extract the s3 logic into a helper?

restapi/stories.go
```diff
@@ -72,6 +80,106 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+	file, h, err := r.FormFile("image")
+	if err != nil {
+		return rest.ErrInvalidRequest(api.logger, "Error getting the image.", err)
+	}
+
+	defer file.Close()
```
AbhiPrasad: Why the defer? It's a long function anyway so can we just close the file after it's uploaded?

restapi/stories.go
```diff
@@ -72,6 +80,106 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+		Body:          fileBytes,
+		Bucket:        bucket,
+		Key:           aws.String(h.Filename),
+		ContentLength: aws.Int64(size),
+		ContentType:   aws.String(fileType),
+	})
+	if err != nil {
+		return rest.ErrInvalidRequest(api.logger, "Error uploading file to s3", err)
+	}
+	resp := "https://" + os.Getenv("BUCKET_NAME") + ".s3.us-west-000.backblazeb2.com/" + h.Filename
```
AbhiPrasad: prefer `fmt.sprintf` over string concatenation with `+`

restapi/stories.go
```diff
@@ -72,6 +80,106 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+			}
+		}
+		return rest.ErrInternal(api.logger, err)
+	}
+
+	story.ImageURL = resp
+	story.Title = r.FormValue("title")
+	story.Content = r.FormValue("content")
+	story.CurrentCity = r.FormValue("current_city")
+	i, err := strconv.ParseUint(r.FormValue("year"), 10, 64)
```
AbhiPrasad: Let's use a more descriptive variable then `i`?

restapi/stories.go
```diff
@@ -72,6 +80,106 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+	story.Summary = r.FormValue("summary")
+	city := story.CurrentCity
+	coordinates, err := api.locationFinder.GetCityCenter(city)
+	if err != nil {
+		return rest.ErrInvalidRequest(api.logger, "Story has an invalid current city", err)
+	}
+	story.Latitude = randomCoords(coordinates.Latitude)
+	story.Longitude = randomCoords(coordinates.Longitude)
+	//regex to find youtube video id
+	re := regexp.MustCompile(`(?:youtube\.com\/(?:[^\/]+\/.+\/|(?:v|e(?:mbed)?)\/|.*[?&]v=)|youtu\.be\/)([^"&?\/\s]{11})`)
```
AbhiPrasad: regex is constant, we can move it outside function.

restapi/stories.go
```diff
@@ -72,6 +80,106 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+	videoURL := r.FormValue("video_url")
+	match := re.FindAllStringSubmatch(videoURL, 2)
```
AbhiPrasad: extract regex logic into helper function

restapi/stories.go
```diff
@@ -72,6 +80,106 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+
+	story.VideoURL = "https://www.youtube.com/embed/" + match[0][1]
```
AbhiPrasad: Move `https://www.youtube.com/embed/` into constant

AbhiPrasad (changes requested):
1st pass review.

If you want anything clarified, DM me, we can hop on a call.

restapi/stories.go
```diff
@@ -72,6 +80,106 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+				return rest.ErrInvalidRequest(api.logger, "Unknown origin country", nil)
+			}
+			if err := api.database.Create(&author).Error; err != nil { //Adding author
+				return rest.ErrInternal(api.logger, err)
+			}
+		}
+		return rest.ErrInternal(api.logger, err)
+	}
+
+	story.ImageURL = resp
```
nimwijetunga: some of these fields can be optional (i.e they may not always appear in the api request) so you will need to make sure this logic still works

nimwijetunga: Optional fields include: author bio, tags and youtube link

restapi/stories.go
nimwijetunga: this logic should be wrapped in an if b/c the link may not always be present

restapi/stories.go
```diff
@@ -72,6 +80,106 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+	resp := "https://" + os.Getenv("BUCKET_NAME") + ".s3.us-west-000.backblazeb2.com/" + h.Filename
+
+	var story models.Story
+	var author models.Author
+	author.FirstName = r.FormValue("author_first_name")
+	author.LastName = r.FormValue("author_last_name")
+	author.OriginCountry = r.FormValue("author_country")
+	author.Bio = r.FormValue("bio")
+
+	errAuthor := api.database.First(&author).Error
```
nimwijetunga, line 144: this might be a common thing to do (i.e find an author given a struct) so maybe create a function in the models file that does this and call it here

restapi/stories.go
```diff
@@ -72,6 +80,106 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
 	return rest.MsgStatusOK("Stories added successfully")
 }
 
+func (api api) CreateStoriesFormData(w http.ResponseWriter, r *http.Request) render.Renderer {
+
+	awsAccessKeyID := os.Getenv("B2_KEY_ID")
```
nimwijetunga: are these keys already in heroku? If not add them please

AbhiPrasad:
Also we should write table tests for this (https://github.com/golang/go/wiki/TableDrivenTests), but we can do that after this PR gets merged in.

restapi/api.go
```diff
@@ -37,7 +37,7 @@ func Router(db *gorm.DB, locationFinder location.LocationFinder) (http.Handler,
 		rest.GetHandler(r, "/story/{storyID}", api.ReturnStoryByID)
 		rest.GetHandler(r, "/authors/origin_countries", api.ReturnAllCountries)
 		rest.PostHandler(r, "/login", api.Login)
-
+		rest.PostHandler(r, "/stories_formdata", api.CreateStoriesFormData)
```
nimwijetunga: Can you put this under PrivateAPI we don't want people to add stories until we finish auth

restapi/stories.go
nimwijetunga: +1

nimwijetunga (changes requested):

restapi/stories.go
Shehryar21: image
you mean this api interface in api.go right?

restapi/stories.go
Shehryar21: Is the Frontend doing any data validation making sure that link is a valid youtube link, or should i check in the backend

restapi/stories.go
AbhiPrasad: yup

Shehryar21: would it be better to just have s3Config as a constant since keyId and SecretKey are used in s3config anyways

restapi/stories.go
Shehryar21: I remember we had a discussion where we might plan to change the filesize. In that case, wouldnt i need to defer it?

restapi/stories.go
AbhiPrasad: yup, agree.

restapi/stories.go
AbhiPrasad: Ah yeah true, I guess there is no harm in keeping it for now.

restapi/stories.go
AbhiPrasad: we should always prioritize validation on the backend, never trust what a client sends you.

Being paranoid >>>>

restapi/stories.go
Shehryar21: why in the models file?

restapi/stories.go
Shehryar21: Some of them werent there (E.g bucket name), I added them

restapi/api.go
Shehryar21: Should I put under Private after Auth is done. Might be easier to test and integrate with Frontend cuz i believe we are removing old auth anyways

AbhiPrasad: wait why are we calling the endpoint `stories_formdata`? Shouldn't we just be POSTing on `stories`?

restapi/stories.go
```diff
@@ -72,6 +83,115 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+		ContentType:   aws.String(fileType),
+	})
+	if err != nil {
+		return "", fmt.Errorf("Could not upload to S3")
+	}
+	return fmt.Sprintf("https://%s.s3.us-west-000.backblazeb2.com/%s", os.Getenv("BUCKET_NAME"), name), nil
+}
+
+func convertYoutubeURL(originalURL string) (string, error) {
+	re := regexp.MustCompile(youtubeRegex)
```
AbhiPrasad, line 117: I think we can make this a constant as well and move it outside of the function to the top of the file.

restapi/stories.go
```diff
@@ -72,6 +83,115 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+	match := re.FindAllStringSubmatch(originalURL, 2)
+	if len(match) == 0 {
+		return "", fmt.Errorf("Invalid Youtube Link")
+	}
+	return youtubeEmbedURL + match[0][1], nil
```
AbhiPrasad: Prefer `fmt.Sprintf`

restapi/stories.go
```diff
@@ -72,6 +83,115 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+}
+
+func (api api) CreateStoriesFormData(w http.ResponseWriter, r *http.Request) render.Renderer {
+
+	file, h, err := r.FormFile("image")
+
+	if err != nil {
+		return rest.ErrInvalidRequest(api.logger, "Error getting the image.", err)
+	}
+	res, err := api.s3upload(file, h.Size, h.Filename)
```
AbhiPrasad: Let's name this `image_url` instead of res.

restapi/stories.go
```diff
@@ -72,6 +83,115 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
 	return rest.MsgStatusOK("Stories added successfully")
 }
 
+func (api api) s3upload(file multipart.File, size int64, name string) (string, error) {
```
AbhiPrasad: nit: Let's call this function `uploadImageTos3`, make it more detailed

restapi/stories.go
```diff
@@ -72,6 +83,115 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+		return rest.ErrInternal(api.logger, err)
+	}
+
+	story.ImageURL = res
+	story.Title = r.FormValue("title")
+	story.Content = r.FormValue("content")
+	story.CurrentCity = r.FormValue("current_city")
+	year, err := strconv.ParseUint(r.FormValue("year"), 10, 64)
+	if err != nil {
+		return rest.ErrInvalidRequest(api.logger, "Error parsing year field", nil)
```
AbhiPrasad:
```suggestion
		return rest.ErrInvalidRequest(api.logger, "Error parsing year field", err)
```

restapi/stories.go
```diff
@@ -72,6 +83,115 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+func convertYoutubeURL(originalURL string) (string, error) {
+	re := regexp.MustCompile(youtubeRegex)
+	match := re.FindAllStringSubmatch(originalURL, 2)
+	if len(match) == 0 {
+		return "", fmt.Errorf("Invalid Youtube Link")
+	}
+	return youtubeEmbedURL + match[0][1], nil
+}
+
+func (api api) CreateStoriesFormData(w http.ResponseWriter, r *http.Request) render.Renderer {
```
AbhiPrasad, line 125: hmm, what happens if we get an array of stories in this endpoint? Shouldn't we be looping through the array?

AbhiPrasad (commented):
Almost there :)

restapi/stories.go
nimwijetunga: i feel like any methods where we query the model given some parameters should live in a specific file so they can be reused (I think the models file is the most sensible unless you can think of somewhere else?)

Shehryar21: How about in the authors.go file since all the code related to authors is in that file?

restapi/api.go
Shehryar21: So i did not want to change the existing endpoint, which is why created a new one. Also the difference between these two is that `stories_formdata` (we can call it smth else or if we decide to get rid of the previous one, we can just call it stories) request payload is a formdata where as `stories` endpoint is a json. Also `stories_formdata` creates a single story, not sure how you would pass in a array of formdata. And looking at our use case, submitting a form on the Frontend would create just one story.

AbhiPrasad: I think we should just make the frontend send as `application/json` instead of form data, and make the endpoints JSON only.

Shehryar21: How would the FE pass in the image file in json (base64 encoding??) not sure how we would convert that in the backend but i can take a look at that if thats what we plan on doing. With form data, FE is just passing the image file directly.

AbhiPrasad: I mean the image file is just a byte array, so it can be accessed the same way in the JSON object iirc. 

I guess if it's too complicated we can leave it as is.

restapi/stories.go
Shehryar21: I dont think regex.MustCompile can be initialized as a constant

restapi/api.go
Shehryar21: `file, h, err := r.FormFile("image")` This is how it reads the file from formdata. If its json, it would not be like this

restapi/stories.go
```diff
@@ -129,7 +129,7 @@ func (api api) CreateStoriesFormData(w http.ResponseWriter, r *http.Request) ren
 	if err != nil {
 		return rest.ErrInvalidRequest(api.logger, "Error getting the image.", err)
 	}
-	res, err := api.s3upload(file, h.Size, h.Filename)
+	ImageURL, err := api.uploadImageTos3(file, h.Size, h.Filename)
```
AbhiPrasad:
```suggestion
	imageURL, err := api.uploadImageTos3(file, h.Size, h.Filename)
```

prefer lowercase here

AbhiPrasad (approved):
lgtm if it works after testing

nimwijetunga (approved):
lgtm (as long as it works after testing with Hanlin)

PR #218, opened 2021-02-08
feat: show popups on pin hover

hanlinc27:
I don't think there's an explicit ticket for this but figma shows that pin behaviour should show popup on hover so this is just a small fix for that. It shouldn't change anything for onclick behaviour. 
d3ffef5d-9498-4152-8118-b4e793d45caa


ui/src/components/Pin.tsx
```diff
@@ -62,7 +62,16 @@ export function Pin({
   const rightPaddingPoint = L.point(100, 400);
 
   return (
-    <Marker position={[story.latitude, story.longitude]} icon={icon}>
+    <Marker
+      position={[story.latitude, story.longitude]}
+      icon={icon}
+      onMouseOver={(e) => {
+        e.target.openPopup();
```
AbhiPrasad, line 69: What happens if someone has a marker open for marker A, what happens when you mouse over marker B?

hanlinc27: It will close marker A's preview since it left that target and open the preview for marker B

hanlinc27:
Closing for now, should implement marker becoming bigger on hover, not preview.

PR #219, opened 2021-02-09
google oauth setup

### LINKED ISSUE #204: Google Login - Backend
- access google console with a blueprint email

- implement logic to detect wether user entered email is a .theshoeproject.online domain
    - .uwblueprint.org email is also valid

- create the user in the backend if they are valid
  - if the email is invalid redirect to an error page
  - if its valid redriect to all stories page (random page for now)
      - generate a JWT token & store _somewhere_ where we can validate it (session cookie?) and return to frontend

s278gupt:
Closes: https://github.com/uwblueprint/shoe-project/issues/204

- Set up project on google api console
- Create `/login` endpoint
  - Sets `state` and redirects to google oauth endpoint
  - Google oauth set up to call our callback endpoint below
- Create `/auth/callback` endpoint
  - Checks state 
  - Checks email domain
  - Retrieves user email
  - Encodes + returns jwt token (setting jwt token in local session storage is for next pr)
- If unauthorised, redirects to vanilla `/unauthorized` endpoint
- Updates some of the frontend boilerplate setup (due to changes in `User` model)

<b>To reviewers:</b>
Feel free to tophat by going to `localhost:8900/api/login` (if you're unable to login to Google you might need to be added to our project on the google console -- please slack me!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ You can check locally to make sure none of these issues show up by:
 make backend-check
 ```
 
-## Authentication
+<!-- ## Authentication
 All the `create` endpoints are protected, which means you have to login before you can make POST requests to these endpoints.
 
 1. Make sure that you have the latest `.env` file (pinned on Slack channel) in the *root* of your `shoe-project` directory.
@@ -145,7 +145,7 @@ $ TOKEN=<JWT-Token>
 $ curl -H 'Accept: application/json' -H "Authorization: Bearer ${TOKEN}" http://localhost:8900/api/<protected-endpoint>
 ```
 
-Note that the key `Authorization: Bearer` is fixed, so don't change this.
+Note that the key `Authorization: Bearer` is fixed, so don't change this. -->
 
 ## Running Script
 
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ You can check locally to make sure none of these issues show up by:
 ```bash
 make backend-check
 ```
-
+<!-- TOOD: add auth docs -->
 <!-- ## Authentication
 All the `create` endpoints are protected, which means you have to login before you can make POST requests to these endpoints.
 
--- a/config/getter.go
+++ b/config/getter.go
@@ -183,6 +183,14 @@ func GetTokenIssuer() string {
 	return viper.GetString("auth.jwt_issuer")
 }
 
+func GetGoogleClientId() string {
+	return viper.GetString("google.client_id")
+}
+
+func GetGoogleClientSecret() string {
+	return viper.GetString("google.client_secret")
+}
+
 func GetMapBoxToken() string {
 	return viper.GetString("mapbox.token")
 }
--- a/config/getter.go
+++ b/config/getter.go
@@ -165,14 +165,6 @@ func GetJWTKey() *jwtauth.JWTAuth {
 	return jwtauth.New("HS256", []byte(viper.GetString("auth.jwt_key")), nil)
 }
 
-func GetSuperUserUsername() string {
-	return viper.GetString("auth.superuser_username")
-}
-
-func GetSuperUserPassword() string {
-	return viper.GetString("auth.superuser_password")
-}
-
 func GetTokenExpiryDuration() (time.Duration, error) {
 	timeProvider := viper.GetString("auth.jwt_expiry")
 
--- a/internal/database/models/auth.go
+++ b/internal/database/models/auth.go
@@ -1,6 +1,8 @@
 package models
 
 import (
+	"errors"
+
 	"github.com/dgrijalva/jwt-go"
 	"gorm.io/gorm"
 )
@@ -11,6 +13,14 @@ type User struct {
 	gorm.Model
 	Email string `json:"email" gorm:"unique; not null"`
 	Hd    string `json:"hd" gorm:"not null"`
+}
+
+// BeforeCreate validates that a user email belongs to either uwblueprint or theshoeproject
+func (user *User) BeforeCreate(tx *gorm.DB) (err error) {
+	if user.Hd != "uwblueprint.org" && user.Hd != "theshoeproject.online" {
+		return errors.New("Invalid email domain")
+	}
+	return nil
 }
 
 // Claims is a custom JWT claim type
--- a/internal/database/models/auth.go
+++ b/internal/database/models/auth.go
@@ -11,8 +11,14 @@ type User struct {
 	Password string `json:"password" gorm:"not null"`
 }
 
+type UserV2 struct {
+	gorm.Model
+	Email string `json:"email" gorm:"unique; not null"`
+	Hd    string `json:"hd" gorm:"not null"`
+}
+
 // Not persisted in db
 type Claims struct {
-	Username string `json:"username"`
+	Email string `json:"email"`
 	jwt.StandardClaims
 }
--- a/internal/database/models/auth.go
+++ b/internal/database/models/auth.go
@@ -5,19 +5,16 @@ import (
 	"gorm.io/gorm"
 )
 
+// User stores all users that have ever signed in with a
+// uwblueprint or theshoeproject domains (stored in Hd field)
 type User struct {
-	gorm.Model
-	Username string `json:"username" gorm:"unique; not null"`
-	Password string `json:"password" gorm:"not null"`
-}
-
-type UserV2 struct {
 	gorm.Model
 	Email string `json:"email" gorm:"unique; not null"`
 	Hd    string `json:"hd" gorm:"not null"`
 }
 
-// Not persisted in db
+// Claims is a custom JWT claim type
+// not persisted in db
 type Claims struct {
 	Email string `json:"email"`
 	jwt.StandardClaims
--- a/restapi/auth.go
+++ b/restapi/auth.go
@@ -19,6 +19,11 @@ import (
 	"gorm.io/gorm"
 )
 
+var oauthconfig *oauth2.Config
+
+// What is state?
+// Since your redirect_uri can be guessed, using a state value can increase your assurance that
+// an incoming connection is the result of an authentication request.
 func generateStateOauthCookie(w http.ResponseWriter) (string, error) {
 	duration, err := config.GetTokenExpiryDuration()
 	if err != nil {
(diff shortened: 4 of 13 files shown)


README.md
```diff
@@ -115,7 +115,7 @@ You can check locally to make sure none of these issues show up by:
 make backend-check
 ```
 
-## Authentication
+<!-- ## Authentication
```
AbhiPrasad, line 118: Let's make sure that we have a TODO somewhere here to add back the Auth docs

restapi/auth.go
```diff
@@ -11,38 +14,99 @@ import (
...
+	cookie := http.Cookie{Name: "oauthstate", Value: state, Expires: expiration, HttpOnly: true} // in prod, set Secure to true
+	http.SetCookie(w, &cookie)
+
+	return state, nil
+}
+
 func (api api) Login(w http.ResponseWriter, r *http.Request) render.Renderer {
-	var user models.User
-	err := json.NewDecoder(r.Body).Decode(&user)
+	oauthconfig = &oauth2.Config{
```
AbhiPrasad: We can't declare this outside the function (maybe we should add it as a field to the api struct)

restapi/auth.go
```diff
@@ -11,38 +14,99 @@ import (
...
+	err = json.NewDecoder(response.Body).Decode(&user)
+	if err != nil {
 		return rest.ErrInternal(api.logger, err)
 	}
 
-	// Validate password
-	if err = bcrypt.CompareHashAndPassword([]byte(dbUser.Password), []byte(user.Password)); err != nil {
-		return rest.ErrUnauthorized(fmt.Sprintf("Wrong password for username %s", user.Username))
+	// redirect if invalid
+	if user.Hd != "uwblueprint.org" && user.Hd != "theshoeproject.online" {
```
AbhiPrasad: Should we make this a struct method on user?

```go
func (user u) hasValidEmail() bool {
    return user.Hd == "uwblueprint.org" || user.Hd == "theshoeproject.online"
}
```

that way if we decide to change `Hd` in the future, we don't have to reach into the API

restapi/endpoints_test.go
```diff
@@ -242,7 +229,7 @@ func (suite *endpointTestSuite) TestCreateAuthorFailsWithUnknownCountry() {
 		},
 	}
 
-	response := suite.endpoint.POST("/authors").WithHeader("Authorization", fmt.Sprintf("Bearer %s", suite.token)).
+	response := suite.endpoint.POST("/authors").
```
AbhiPrasad, line 306: Wait the endpoint doesn't need a header anymore?

restapi/auth.go
```diff
@@ -11,38 +14,99 @@ import (
...
+	if err != nil {
+		return "", err
+	}
+	var expiration = time.Now().Add(duration)
+
+	b := make([]byte, 16)
+	rand.Read(b)
+	state := base64.URLEncoding.EncodeToString(b)
+	fmt.Println(state)
+	cookie := http.Cookie{Name: "oauthstate", Value: state, Expires: expiration, HttpOnly: true} // in prod, set Secure to true
```
AbhiPrasad: > in prod, set Secure to true

Can we do this by looking at the ENV variables for the time being?

restapi/auth.go
```diff
@@ -11,38 +14,99 @@ import (
...
+	duration, err := config.GetTokenExpiryDuration()
+	if err != nil {
+		return "", err
+	}
+	var expiration = time.Now().Add(duration)
+
+	b := make([]byte, 16)
+	rand.Read(b)
+	state := base64.URLEncoding.EncodeToString(b)
+	fmt.Println(state)
```
AbhiPrasad: can we `rm -rf` the println?

internal/database/models/auth.go
```diff
@@ -5,14 +5,17 @@ import (
...
 )
 
+// User stores all users that have ever signed in with a
+// uwblueprint or theshoeproject domains (stored in Hd field)
 type User struct {
 	gorm.Model
-	Username string `json:"username" gorm:"unique; not null"`
-	Password string `json:"password" gorm:"not null"`
+	Email string `json:"email" gorm:"unique; not null"`
+	Hd    string `json:"hd" gorm:"not null"`
```
AbhiPrasad, line 15: Hmm I'm wondering if we can write a CHECK constraint here to limit the Hd values to one of the domains, but idk how hard it is with gorm.

restapi/endpoints_test.go
s278gupt: No, I moved the endpoints around so this is no longer protected (temporarily) and the tests are updated accordingly so they don't fail (the tests will be all updated once at the end of this auth epic)

restapi/auth.go
s278gupt: good call

restapi/auth.go
s278gupt: i actually think it's the opposite...if things change in the future, we don't want to reach into database-layer logic? @nimwijetunga what do you think?

restapi/auth.go
```diff
@@ -11,38 +13,85 @@ import (
...
+
+	url := api.oauthconfig.AuthCodeURL(state)
+	http.Redirect(w, r, url, http.StatusTemporaryRedirect)
+
+	return rest.JSONStatusOK(url)
+}
+
+func (api api) AuthCallback(w http.ResponseWriter, r *http.Request) render.Renderer {
+	expectedState, _ := r.Cookie("oauthstate")
+	if expectedState.Value != r.FormValue("state") {
```
nimwijetunga, line 49: Just so I understand the `r.FormValue("state")` is added on by google right? We don't have to include anything on our frontend?

restapi/auth.go
nimwijetunga: Well I actually think @AbhiPrasad suggestion makes sense. Since this logic is custom to us (i.e has nothing to do with the actual google OAUTH flow) it would be better to make it a member of the User model to encapsulate that logic.

restapi/auth.go
nimwijetunga: Same for `r.FormValue("code")` below right?

restapi/auth.go
```diff
@@ -11,38 +13,85 @@ import (
...
 
-	jwtExpiry, err := config.GetTokenExpiryDuration()
+	// if valid, save to db and create jwt token
+	api.database.FirstOrCreate(&user)
+	jwtToken, err := generateJWTToken(user.Email)
 	if err != nil {
 		return rest.ErrInternal(api.logger, err)
 	}
 
+	return rest.JSONStatusOK(jwtToken)
```
nimwijetunga, line 84: I thought we were going to put the JWT token in the cookie of the response?

nimwijetunga: Also we should redirect users to the all stories page if the authentication is successful. Since this page currently doesn't exist we can redirect them to `/` (i.e homepage) and create another (separate) pr to change this once all stories has been implemented.

restapi/auth.go
```diff
@@ -11,38 +13,85 @@ import (
...
+	if err != nil {
 		return rest.ErrInternal(api.logger, err)
 	}
 
-	// Validate password
-	if err = bcrypt.CompareHashAndPassword([]byte(dbUser.Password), []byte(user.Password)); err != nil {
-		return rest.ErrUnauthorized(fmt.Sprintf("Wrong password for username %s", user.Username))
+	// redirect if invalid
+	if user.Hd != "uwblueprint.org" && user.Hd != "theshoeproject.online" {
+		http.Redirect(w, r, "/api/unauthorized", http.StatusTemporaryRedirect)
```
nimwijetunga, line 75: This should be `/unauthorized` (i.e remove the api prefix). This page should exist when [this pr](https://github.com/uwblueprint/shoe-project/pull/223) is merged in.

restapi/redirects.go
```diff
@@ -0,0 +1,12 @@
+package restapi
```
nimwijetunga, line 1: We don't need this since we can redirect directly to `/unauthorized` which is a frontend page

nimwijetunga (changes requested):
Just a few small things to fix, looks good overall!

restapi/auth.go
s278gupt: ok, this will be done in next pr! didn't want to touch jwt flow in the same pr...

restapi/auth.go
s278gupt: will be changed in next pr (for jwt flow) so I don't have to wait for Dinu's pr to be merged.

restapi/auth.go
s278gupt: yep, this callback endpoint is only ever called by google; all the query params are set by them too

internal/database/models/auth.go
s278gupt: see above

restapi/auth.go
s278gupt: Ok, I've added this logic in a hook before any user record is created (see latest commit)

config/getter.go
```diff
@@ -183,6 +175,14 @@ func GetTokenIssuer() string {
 	return viper.GetString("auth.jwt_issuer")
 }
 
+func GetGoogleClientId() string {
```
nimwijetunga, line 178: Can you add these to heroku as well otherwise we might forget

restapi/auth.go
```diff
@@ -11,38 +13,85 @@ import (
...
+	// exchange received authorization code for an access token
+	token, err := api.oauthconfig.Exchange(context.Background(), r.FormValue("code"))
+	if err != nil {
+		return rest.ErrUnauthorized("Invalid authorization code")
+	}
+
+	// use access token to create a client which we use to access user info
+	user := models.User{}
+	client := api.oauthconfig.Client(context.Background(), token)
+	response, err := client.Get("https://www.googleapis.com/oauth2/v3/userinfo")
```
nimwijetunga, line 62: Maybe we should extract this URL into a constant?

nimwijetunga (approved):
LGTM! Please make sure you test this to see if the flow works before merging!

AbhiPrasad (approved):
Make sure we run `go mod tidy` and `make backend-fmt` before merge

config/getter.go
s278gupt: https://app.zenhub.com/workspaces/the-shoe-project-f20-5f729b8f69fd6b2517b9a848/issues/uwblueprint/shoe-project/243

PR #221, opened 2021-02-10
Pin placement

aspensantos:


restapi/stories.go
```diff
@@ -46,6 +52,16 @@ func (api api) ReturnStoryByID(w http.ResponseWriter, r *http.Request) render.Re
 	return rest.JSONStatusOK(story)
 }
 
+func (api api) randomizePinLocation(city string) (latitude float64, longitude float64, err error) {
```
AbhiPrasad: feels kinda icky that this is an `api` struct method, but I guess it's fine

restapi/stories.go
```diff
@@ -57,12 +73,74 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
 	for i := 0; i < len(stories); i++ {
 		city := stories[i].CurrentCity
-		coordinates, err := api.locationFinder.GetCityCenter(city)
+		stories[i].CurrentCity = strings.Title(strings.ToLower(city))
+
+		// get number of stories with current city in db
+		var prevStories []models.Story
+		err := api.database.Where("current_city=?", city).Find(&prevStories).Error
+		if err != nil {
+			api.logger.Infof("A")
```
AbhiPrasad: More descriptive log message? (Ditto with stuff below it)

restapi/stories.go
```diff
@@ -57,12 +73,74 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+			}
+			continue
+		}
+		numStoriesInCity := 0
+		if prevStories != nil {
+			numStoriesInCity = len(prevStories)
+		}
+		// query zipcode with numStoriesInCity + 1 results and take the last one
+		zipCodeToken := os.Getenv("ZIP_CODE_TOKEN")
+		resp, err := http.Get(fmt.Sprintf("https://app.zipcodebase.com/api/v1/code/city?apikey=%s&city=%s&country=ca&limit=%d", zipCodeToken, city, numStoriesInCity+1))
```
AbhiPrasad: Extract URL into constant?

restapi/stories.go
```diff
@@ -57,12 +73,74 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+				return rest.ErrInvalidRequest(api.logger, fmt.Sprintf("Story %d has an invalid current city", i), err)
+			}
+			continue
+		}
+		numStoriesInCity := 0
+		if prevStories != nil {
+			numStoriesInCity = len(prevStories)
+		}
+		// query zipcode with numStoriesInCity + 1 results and take the last one
+		zipCodeToken := os.Getenv("ZIP_CODE_TOKEN")
```
AbhiPrasad: Move out of function into constant?

restapi/stories.go
```diff
@@ -57,12 +73,74 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+			api.logger.Infof("D")
+			// default to randomizing if there is an error
+			stories[i].Latitude, stories[i].Longitude, err = api.randomizePinLocation(city)
+			if err != nil {
+				return rest.ErrInvalidRequest(api.logger, fmt.Sprintf("Story %d has an invalid current city", i), err)
+			}
+			continue
+		}
+
+		postalCode := zipcodeResp.Results[len(zipcodeResp.Results)-1]
```
AbhiPrasad: Why the last element of the array?

AbhiPrasad (commented):
1st pass, DM me if you got questions.

restapi/stories.go
nimwijetunga: Also when logging errors please use `logger.Fatalw("...")`

restapi/stories.go
```diff
@@ -57,12 +73,74 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
 
 	for i := 0; i < len(stories); i++ {
 		city := stories[i].CurrentCity
-		coordinates, err := api.locationFinder.GetCityCenter(city)
+		stories[i].CurrentCity = strings.Title(strings.ToLower(city))
```
nimwijetunga: Can we move this logic out into a function? I would like the create stories endpoint to be clean

restapi/stories.go
```diff
@@ -57,12 +73,74 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+
+		// get number of stories with current city in db
+		var prevStories []models.Story
```
nimwijetunga, line 74: I think in gorm you can use `.count` (or something similar) to get the count. I think we should use that as loading all stories in memory (even if its a single city) could cause us issues. [https://stackoverflow.com/questions/54720000/how-display-result-count-from-query](https://stackoverflow.com/questions/54720000/how-display-result-count-from-query)

restapi/stories.go
```diff
@@ -57,12 +73,74 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
+		err := api.database.Where("current_city=?", city).Find(&prevStories).Error
+		if err != nil {
+			api.logger.Infof("A")
+			// default to randomizing if there is an error
+			stories[i].Latitude, stories[i].Longitude, err = api.randomizePinLocation(city)
```
nimwijetunga: You seem to default to randomizing in a few places so its best to extract this into a helper function

restapi/stories.go
nimwijetunga: Esentially the number of postal codes we request from the API is how many stories we currently have + 1 (for the new story). So we get the last postal code because it is the newest (the API always returns postal codes in the same order so we can't request 1 each time otherwise we would get the same one).

Also we can't load all postal codes each time otherwise we'd hit the api limit in 2-3 api requests.

nimwijetunga (changes requested):

restapi/stories.go
aspensantos: moved it

restapi/stories.go
aspensantos: got rid of all the logs

restapi/stories.go
aspensantos: done

restapi/stories.go
aspensantos: done

restapi/stories.go
aspensantos: moved it all into a function

restapi/stories.go
aspensantos: okay used count

restapi/stories.go
aspensantos: moved

internal/location/coordinates.go
```diff
@@ -0,0 +1,112 @@
...
+	return Coordinates{
+		Latitude:  forward.Features[0].Center[1],
+		Longitude: forward.Features[0].Center[0],
+	}, nil
+}
+
+func randomCoords(coordinate float64) float64 {
+	num := math.Floor(rand.Float64()*99) + 1
+	if math.Round(rand.Float64()) == 1 {
+		num *= 1
```
AbhiPrasad, line 69: we multiply the number... by 1? lol

internal/location/coordinates.go
```diff
@@ -0,0 +1,112 @@
...
+	latitude = randomCoords(coordinates.Latitude)
+	longitude = randomCoords(coordinates.Longitude)
+	return latitude, longitude, nil
+}
+
+func (finder coordinatesFinder) GetLatitudeAndLongitude(city string, limit int64) (float64, float64, error) {
+	// query zipcode with numStoriesInCity + 1 results and take the last one
+	resp, err := http.Get(fmt.Sprintf(zipCodeURL, finder.zipCodeToken, city, limit+1))
+	if err != nil || resp.StatusCode != 200 {
+		// default to randomizing if there is an error
```
nimwijetunga, line 90: It might be useful in these cases to do some logging (i.e `log.fatalw(...)`) so we can trace things down easier and know when it fails

nimwijetunga: I suggest you log each time we default to randomizing (make it a different message for each one so we know where it fails)

internal/location/coordinates.go
```diff
@@ -0,0 +1,112 @@
...
+	coordinates, err := finder.GetCityCenter(city)
+	if err != nil {
+		return 0, 0, err
+	}
+	latitude = randomCoords(coordinates.Latitude)
+	longitude = randomCoords(coordinates.Longitude)
+	return latitude, longitude, nil
+}
+
+func (finder coordinatesFinder) GetLatitudeAndLongitude(city string, limit int64) (float64, float64, error) {
```
nimwijetunga: Rename function to `GetPostalLatitudeAndLongitude`

restapi/stories.go
```diff
@@ -67,13 +66,17 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
...
 	for i := 0; i < len(stories); i++ {
-		city := stories[i].CurrentCity
-		coordinates, err := api.locationFinder.GetCityCenter(city)
+		stories[i].CurrentCity = strings.Title(strings.ToLower(stories[i].CurrentCity))
+
+		// get number of stories with current city in db
+		var prevStories []models.Story
+		var numStoriesInCity int64
+		var err error
+		api.database.Where("current_city=?", stories[i].CurrentCity).Model(&prevStories).Count(&numStoriesInCity)
```
nimwijetunga: Something to note: You should manually increment the numStoriesInCity field each iteration after the first query for two reasons

1. You don't need to query the database multiple times (chances are very low we have concurrent writes to the database so this should be fine)
2. Doing the query is inaccurate because you never add the story to the database until you've looped over all stories. So the count will return the same number each time

restapi/stories.go
```diff
@@ -158,20 +161,22 @@ func (api api) CreateStoriesFormData(w http.ResponseWriter, r *http.Request) ren
 	story.ImageURL = imageURL
 	story.Title = r.FormValue("title")
 	story.Content = r.FormValue("content")
-	story.CurrentCity = r.FormValue("current_city")
+	story.CurrentCity = strings.Title(strings.ToLower(r.FormValue("current_city")))
```
nimwijetunga, line 173: I would make sure this doesn't break any existing functionality (i.e do a quick search in the repo for the `currentCity` usages). If it does then let's just do the conversion in the function call `GetLatitudeAndLongitude`

internal/location/coordinates.go
```diff
@@ -0,0 +1,112 @@
...
+	"math/rand"
+	"net/http"
+	"os"
+
+	mapbox "github.com/ryankurte/go-mapbox/lib"
+	"github.com/ryankurte/go-mapbox/lib/geocode"
+)
+
+const geocodeRequestLimit = 1
+const zipCodeTokenKey = "ZIP_CODE_TOKEN"
```
nimwijetunga: Can we add this and any other new env variables (that you used for the postal code API) to Heroku. If you don't know how to do this we can take a look next work session

nimwijetunga (changes requested):
Almost there! Just a couple of things to fix

restapi/stories.go
AbhiPrasad: Should we be also using https://gorm.io/docs/transactions.html#Transaction here?

nimwijetunga: Hmm yeah using a transaction seems cleaner

internal/location/coordinates.go
aspensantos: lol idk i didnt write this just moved it from the other file

internal/location/coordinates.go
aspensantos: added logs

restapi/stories.go
aspensantos: checked our current tests and it does not break them

restapi/stories.go
aspensantos: Used a dictionary since we cannot just increment one value as the city could be different on each iteration, and reading what you update in a transaction gives you the outdated value

internal/location/coordinates.go
aspensantos: I changed the token stuff to match how we did it for mapbox

nimwijetunga (approved):
LGTM please test this before merging!

PR #222, opened 2021-02-10
Nuke DB Tables Before Seeding

nimwijetunga:
Otherwise we get errors that data already exists and/or end up with duplicates


AbhiPrasad (approved):

AbhiPrasad:
lol drop table development lets go!!!!

PR #223, opened 2021-02-10
Login page

DinuWije:
## Functionality:
Currently logs a message to console and redirects the user to /admin endpoint when Google Sign In button is clicked

## Todo:
- Add API call to backend to log the user in when the button is clicked
- Edit CSS once the Figma designs are hi-fi
## /login endpoint: 
image

## /unauthorized endpoint:
image

### DIFF
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -26,7 +26,7 @@ function App(): JSX.Element {
             <Route exact path="/">
               <ShoeMap />
             </Route>
-            <Route path="/login">
+            <Route exact path="/login">
               <Login />
             </Route>
             <PrivateRoute path="/admin">
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -27,7 +27,10 @@ function App(): JSX.Element {
               <ShoeMap />
             </Route>
             <Route exact path="/login">
-              <Login />
+              <Login login={true} />
+            </Route>
+            <Route exact path="/unauthorized">
+              <Login login={false} />
             </Route>
             <PrivateRoute path="/admin">
               <Admin />
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -27,7 +27,10 @@ function App(): JSX.Element {
               <ShoeMap />
             </Route>
             <Route exact path="/login">
-              <Login />
+              <Login login />
+            </Route>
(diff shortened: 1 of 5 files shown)


AbhiPrasad (approved):
sick, good stuff.

Plz run `yarn fix` before merge!

ui/src/pages/Login.tsx
```diff
@@ -1,14 +1,76 @@
...
+  position: absolute;
+`;
+
+const StyledCard = styled(Card)`
+  max-height: 35vh;
+  max-width: 25vw;
+`;
+
+const TitleText = styled.p`
+  font-size: 20px;
```
hanlinc27: can you move the TitleText and Message Text to the typography file? And then use fontSize.whicheverText for consistency.

hanlinc27: Wait nvm I realized this is all lofi so the font sizes are still in the air - would be good though to still keep the message and title texts in the typography file like the other components :)

ui/src/pages/Login.tsx
```diff
@@ -1,14 +1,76 @@
 import * as React from "react";
 import { Redirect } from "react-router-dom";
```
hanlinc27: run `yarn run fix`

ui/src/App.tsx
```diff
@@ -26,8 +26,11 @@ function App(): JSX.Element {
             <Route exact path="/">
               <ShoeMap />
             </Route>
-            <Route path="/login">
-              <Login />
+            <Route exact path="/login">
+              <Login login={true} />
```
hanlinc27: it will have a value of **true** by default so you can just put **login** instead of explicitly defining

hanlinc27 (changes requested):

ui/src/pages/Login.tsx
```diff
@@ -23,29 +22,22 @@ const StyledCard = styled(Card)`
...
   text-align: center;
   margin-bottom: 5vh;
 `;
 
-export const Login: React.FC<{ login: boolean }> = ({ login }) => {
+interface LoginProps {
+  login: boolean;
+}
+
+export const Login: React.FC<LoginProps> = (props) => {
```
AbhiPrasad: let's just destructure the props here

hanlinc27 (approved):

PR #225, opened 2021-02-13
fix: PinPreview small styling + Progressive Image Loading

### LINKED ISSUE #189: Small fixes to preview
- fix box shadow/rounded corner issue of top corners (see image)
- add loading circle animation, it takes a while for the shoe image to show up

AbhiPrasad:
Fixes #189

We do progressive image loading using a double image method. We initially render a loading indicator. We then load the image through `display: none;` img element. When that hidden img triggers `onLoad` event, we render the image with a visible img element.

Fixed small styling stuff to pin preview to make it more like designs.

image

Kapture 2021-02-13 at 16 09 46

### DIFF
--- a/ui/src/components/StoryDrawer.tsx
+++ b/ui/src/components/StoryDrawer.tsx
@@ -55,7 +55,6 @@ const LoadingContent = styled.div`
   display: flex;
   align-items: center;
   justify-content: center;
-  width: 100%;
   height: 500px;
 
   @media ${device.mobileS} {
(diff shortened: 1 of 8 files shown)


ui/src/components/StoryDrawer.tsx
```diff
@@ -50,6 +51,19 @@ const StyledImage = styled.img`
   }
 `;
 
+const LoadingContent = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 100%;
```
hanlinc27: Don't think we need width: 100% since it's already constrained by the Grid. 
```suggestion
```

hanlinc27 (approved):

Issue #226, opened 2021-02-13
All Stories - Backend


nimwijetunga:
- Modify the stories model to have a column which indicates whether the story is visible or not (i.e a boolean) on the map
     - **Note:** This blocks frontend so get a pr done for this ASAP and get it merged (separate from the other tasks in this ticket)
- Create an endpoint that returns all stories in the backend (may be able to just use the current all stories endpoint)
- Modify the all stories endpoint to accept a query parameter: `tags=[tagA, tagB, tagC]`  which gets all the stories that have been tagged with fields in the `tag` array (as long as a story has at least one of the tags in tag array you should return it)
- Modify the all stories endpoint to accept a query parameter: `visibility=[true/false]` which will get stories that are visible or hidden
- Modify the all stories endpoint to accept two query parameters: `sort=[columnA, columnB, ...]&order=[asc,desc,...]` where you return all the stories sorted by the respective columns in the appropriate order. For the example above sort the stories on columnA ascending and columnB descending
   - **Note:** Sorting should be applied after filtering and should be the last step in the endpoint before returning the stories
   - The sort columns are: `story_name, year, author_name, author_country`

DinuWije:
To send arrays as query parameters do the following:
arrayName?=arrayVal1&arrayName=arrayVal2& ... 

Example:
localhost:8900/api/stories?sort=current_city&sort=year&order=asc&order=desc&tags=EDUCATION&tags=REFUGEE

PR #231, opened 2021-02-17
feat: Add is_visible field to story model

AbhiPrasad:
#226 

Plz tell me if I missed something

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,6 @@ require (
 	github.com/mattn/go-sqlite3 v1.14.4 // indirect
 	github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e // indirect
 	github.com/onsi/ginkgo v1.13.0 // indirect
-	github.com/pkg/errors v0.9.1 // indirect
 	github.com/ryankurte/go-mapbox v0.4.4
 	github.com/snowzach/certtools v1.0.2
 	github.com/spf13/afero v1.4.1 // indirect
--- a/internal/database/migrations/migrations.go
+++ b/internal/database/migrations/migrations.go
@@ -81,7 +81,7 @@ func Seed(db *gorm.DB, locationFinder location.LocationFinder) error {
 		stories[i].Latitude = coordinates.Latitude
 		stories[i].Longitude = coordinates.Longitude
 		// Make sure all stories are public for initially seeded stories
-		stories[i].IsPublic = true
+		stories[i].IsVisible = true
 	}
 
 	return db.Create(&stories).Error
--- a/internal/database/migrations/migrations.go
+++ b/internal/database/migrations/migrations.go
@@ -16,15 +16,15 @@ func CreateTables(db *gorm.DB) error {
 }
 
 func DropAllTables(db *gorm.DB) error {
-	err := db.Migrator().DropTable(&models.User{});
+	err := db.Migrator().DropTable(&models.User{})
 	if err != nil {
-		return err;
+		return err
 	}
-	err = db.Migrator().DropTable(&models.Story{});
+	err = db.Migrator().DropTable(&models.Story{})
 	if err != nil {
-		return err;
+		return err
 	}
-	return db.Migrator().DropTable(&models.Author{});
+	return db.Migrator().DropTable(&models.Author{})
 }
 
 func CreateSuperUser(db *gorm.DB) error {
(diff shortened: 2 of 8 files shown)


internal/database/migrations/migrations.go
```diff
@@ -16,15 +16,15 @@ func CreateTables(db *gorm.DB) error {
 }
 
 func DropAllTables(db *gorm.DB) error {
-	err := db.Migrator().DropTable(&models.User{});
+	err := db.Migrator().DropTable(&models.User{})
```
nimwijetunga, line 19: lol i've been doing too much c++

restapi/stories.go
```diff
@@ -190,6 +189,10 @@ func (api api) CreateStoriesFormData(w http.ResponseWriter, r *http.Request) ren
 	if err := api.database.Create(&story).Error; err != nil {
 		return rest.ErrInternal(api.logger, err)
 	}
+
+	// We set the story to be private when it is initially uploaded
+	story.IsPublic = false
```
DinuWije: What is the point of having a "is_public" json field if you're going to set it to false anyways when you first upload it?

internal/database/models/models.go
```diff
@@ -17,12 +18,14 @@ type Author struct {
...
 }
 
+// Story struct describes a story rendered on the map
 type Story struct {
 	gorm.Model
 	Title           string  `gorm:"not null" json:"title"`
 	Content         string  `gorm:"type:text;not null" json:"content"`
 	CurrentCity     string  `gorm:"not null" json:"current_city"`
 	Year            uint    `gorm:"not null" json:"year"`
+	IsPublic        bool    `gorm:"not null" json:"is_public"`
```
nimwijetunga: Can we rename the field to `IsVisible` (since its a visibility toggle)?

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/uwblueprint/shoe-project
 go 1.14
 
 require (
+	github.com/aws/aws-sdk-go v1.37.12
```
nimwijetunga: why do we need the aws sdk?

nimwijetunga (approved):
Looks good just some small things to address

DinuWije (approved):
We also use the Stories model in the endpoints_test.go file. Would we need to update this file as well?

internal/database/models/models.go
Shehryar21: I think we can set a default value as false `gorm:"default:false"`, rather than setting story.isPublic to False later on

internal/database/migrations/migrations.go
```diff
@@ -81,6 +80,8 @@ func Seed(db *gorm.DB, locationFinder location.LocationFinder) error {
 		}
 		stories[i].Latitude = coordinates.Latitude
 		stories[i].Longitude = coordinates.Longitude
+		// Make sure all stories are public for initially seeded stories
+		stories[i].IsPublic = true
```
Shehryar21, line 84: Dont think we would need to this if we set the default value as false in the gorm model (check my other comment)

restapi/stories.go
Shehryar21: Dont think we would need to this if we set the default value as false in the gorm model (check my other comment)

AbhiPrasad: It'll be private on upload, and only set to public when we publish the map.

go.mod
AbhiPrasad: `go mod tidy` caused this.

Maybe because of the s3 stuff?

restapi/stories.go
```diff
@@ -190,6 +189,10 @@ func (api api) CreateStoriesFormData(w http.ResponseWriter, r *http.Request) ren
 	if err := api.database.Create(&story).Error; err != nil {
 		return rest.ErrInternal(api.logger, err)
 	}
+
+	// We set the story to be public when it is initially uploaded
+	story.IsVisible = true
```
Shehryar21: I thought the story which is uploaded through the form is not public

AbhiPrasad: word, yeah good catch

AbhiPrasad: We can remove that cause we have the gorm default.

nimwijetunga (approved):
Once you filter all stories by isVisible=true it should be good to merge

Shehryar21 (approved):

PR #232, opened 2021-02-20
Tags backend

### LINKED ISSUE #224: Tags Backend
- Create a tags table to establish an association between a tag and story
- schema: name, story_id

- Endpoints required:
1) Get all unique tags
2) Delete tags by story id
3) Get tags by story id

FE would pass array of tags name and backend would add rows in the tags table accordingly.

Shehryar21:
closes #224

- Create a tags table
- Seed tags
- Added a get all unique tags endpoint
- This is what the response payload looks like now
image

- Update create story endpoint to add tags


internal/database/migrations/migrations.go
```diff
@@ -72,16 +77,29 @@ func Seed(db *gorm.DB, locationFinder location.LocationFinder) error {
 	if err != nil {
 		return err
 	}
-
+	var tags []models.Tag
```
AbhiPrasad: We can predeclare the length of the array, no need to use `append` (instead we can just do array indexing)

internal/database/migrations/migrations.go
```diff
@@ -72,16 +77,29 @@ func Seed(db *gorm.DB, locationFinder location.LocationFinder) error {
...
 	// set lat long for coordinates
 	for i, story := range stories {
 		coordinates, err := locationFinder.GetCityCenter(story.CurrentCity)
 		if err != nil {
 			return err
 		}
+		tagChoice := []string{"Education", "Refugee", "Immigration"}
+		randomIndex := rand.Intn(len(tagChoice))
+		pick := tagChoice[randomIndex]
```
AbhiPrasad: Extract into function plz.

internal/database/migrations/migrations.go
```diff
@@ -72,16 +77,29 @@ func Seed(db *gorm.DB, locationFinder location.LocationFinder) error {
...
+		var tag models.Tag
```
AbhiPrasad: We can do inline declare here:

```go
tag := models.Tag{
  Name: pick,
  Story: story,
}
tags = append(tags, tag)
```

restapi/stories.go
```diff
@@ -190,9 +190,26 @@ func (api api) CreateStoriesFormData(w http.ResponseWriter, r *http.Request) ren
 	if err := api.database.Create(&story).Error; err != nil {
 		return rest.ErrInternal(api.logger, err)
 	}
+	fmt.Println(r.FormValue("bio"))
```
AbhiPrasad: 🔥 
```suggestion
```

restapi/stories.go
```diff
@@ -190,9 +190,26 @@ func (api api) CreateStoriesFormData(w http.ResponseWriter, r *http.Request) ren
...
+	names := r.Form["tags"]
+	err = api.AddTags(names, story)
+	if err != nil {
+		return rest.ErrInvalidRequest(api.logger, "Error Adding Tags", err)
+	}
 	return rest.MsgStatusOK("Story Added Successfully")
 }
 
+func (api api) AddTags(names []string, story models.Story) error {
+	var tags []models.Tag
```
AbhiPrasad: Again we can predeclare and just array index to add tags, no need to `append` if not necessary.

restapi/stories.go
```diff
@@ -190,9 +190,26 @@ func (api api) CreateStoriesFormData(w http.ResponseWriter, r *http.Request) ren
...
+	for _, t := range names {
+		var tag models.Tag
```
AbhiPrasad: ditto as above, prefer inline declare here.

AbhiPrasad (changes requested):

internal/database/models/models.go
```diff
@@ -33,3 +33,10 @@ type Story struct {
 	AuthorCountry   string  `json:"author_country"`
 	Author          Author  `gorm:"foreignKey:AuthorFirstName,AuthorLastName,AuthorCountry;references:FirstName,LastName,OriginCountry" json:"author"`
 }
+
+type Tag struct {
+	gorm.Model
+	Name    string `gorm:"not null" json:"name"`
```
nimwijetunga: You should specify the primary key of this table as (Name, StoryID). This will prevent duplicate story tag pairs

restapi/stories.go
nimwijetunga: remove

restapi/stories.go
```diff
@@ -190,9 +190,26 @@ func (api api) CreateStoriesFormData(w http.ResponseWriter, r *http.Request) ren
...
+	fmt.Println(r.FormValue("bio"))
+	names := r.Form["tags"]
+	err = api.AddTags(names, story)
+	if err != nil {
+		return rest.ErrInvalidRequest(api.logger, "Error Adding Tags", err)
+	}
 	return rest.MsgStatusOK("Story Added Successfully")
 }
 
+func (api api) AddTags(names []string, story models.Story) error {
```
nimwijetunga, line 202: consider making this a helper method instead of a method on api

restapi/tags.go
```diff
@@ -0,0 +1,20 @@
...
+
+	"github.com/go-chi/render"
+	"github.com/uwblueprint/shoe-project/internal/database/models"
+	"github.com/uwblueprint/shoe-project/restapi/rest"
+)
+
+func (api api) ReturnAllUniqueTags(w http.ResponseWriter, r *http.Request) render.Renderer {
+	var tags []models.Tag
+
+	err := api.database.Distinct("name").Find(&tags).Error
```
nimwijetunga, line 13: For right now you can return just the tag names (i.e dont need stories). It doesn't make sense to return stories as several stories can have the same tag

nimwijetunga (changes requested):
First pass, looks good few small things to address

restapi/stories.go
Shehryar21: isnt it better to have it as a method on api, because otherwise i would have to pass in db in parameters of the function??

nimwijetunga: Oh i didn't notice you use db, then its fine to keep it in api

restapi/stories.go
```diff
@@ -190,9 +190,29 @@ func (api api) CreateStoriesFormData(w http.ResponseWriter, r *http.Request) ren
...
+	var length = len(names)
+	tags := make([]models.Tag, length)
+	for i, t := range names {
+		tag := models.Tag{
+			Name:    strings.ToUpper(t),
+			StoryID: story.ID,
+		}
+		tags[i] = tag
```
AbhiPrasad: honestly we can probably than just do:
```suggestion
		tags[i] = models.Tag{
			Name:    strings.ToUpper(t),
			StoryID: story.ID,
		}
```

restapi/stories.go
```diff
@@ -190,9 +190,29 @@ func (api api) CreateStoriesFormData(w http.ResponseWriter, r *http.Request) ren
...
+		if err != nil {
+			return rest.ErrInvalidRequest(api.logger, "Error Adding Tags", err)
+		}
+	}
 	return rest.MsgStatusOK("Story Added Successfully")
 }
 
+func (api api) AddTags(names []string, story models.Story) error {
+	var length = len(names)
+	tags := make([]models.Tag, length)
```
AbhiPrasad: no need to declare the var if we are only using it one place 
```suggestion
	tags := make([]models.Tag, len(names))
```

restapi/tags.go
```diff
@@ -0,0 +1,24 @@
...
+	var tags []models.Tag
+	err := api.database.Distinct("name").Find(&tags).Error
+	if err != nil {
+		return rest.ErrInternal(api.logger, err)
+	}
+
+	var length = len(tags)
+	names := make([]string, length)
+	for i, t := range tags {
+		names[i] = t.Name
```
AbhiPrasad, line 19: Aside: holy we really need generics for `map`, `filter` and `reduce` in this language.

AbhiPrasad (approved):
Can we write a test for the endpoint plz?

internal/database/migrations/migrations.go
```diff
@@ -51,6 +56,13 @@ func parseJson(filename string, obj interface{}) error {
 	return nil
 }
 
+func ChooseRandomTag() string {
+	tagChoice := []string{"EDUCATION", "REFUGEE", "IMMIGRATION"}
+	randomIndex := rand.Intn(len(tagChoice))
```
AbhiPrasad, line 61: oh ya, do we have to seed before usage?

restapi/tags.go
```diff
@@ -0,0 +1,24 @@
...
+
+func (api api) ReturnAllUniqueTags(w http.ResponseWriter, r *http.Request) render.Renderer {
+	var tags []models.Tag
+	err := api.database.Distinct("name").Find(&tags).Error
+	if err != nil {
+		return rest.ErrInternal(api.logger, err)
+	}
+
+	var length = len(tags)
+	names := make([]string, length)
```
nimwijetunga: `make([]string, len(tags))`

restapi/stories.go
nimwijetunga: `make([]models.Tag, len(names))`

internal/database/migrations/migrations.go
Shehryar21: yeah or add story through form which has tags

nimwijetunga (approved):
LGTM! Please make sure you test it before merging!

PR #237, opened 2021-02-24
All stories backend

### LINKED ISSUE #226: All Stories - Backend

DinuWije:
## Closes #226 

Check out the ticket for the in-depth functionality of the code in this PR. Here's an overview: 
- filtering by tags 
- filtering by visibility
- sorting based on various fields 

Sample query: 
localhost:8900/api/stories?sort=current_city&sort=year&order=asc&order=desc&tags=EDUCATION&tags=REFUGEE


restapi/stories.go
```diff
@@ -34,8 +34,54 @@ func init() {
...
+
+	err := gormTags.Where("name IN ?", tags).Find(&storiesByTags).Error
+	if err != nil {
+		return rest.ErrInternal(api.logger, err)
+	}
+
+	if len(storiesByTags) != 0 {
+		var storyIDs []string
+		for i:=0; i<len(storiesByTags); i++{
+			storyIDs = append(storyIDs, fmt.Sprint(storiesByTags[i].StoryID))
```
AbhiPrasad: We don't need to append every time, we already know the final length of the storyIDs array a head of time. We can just array index.

Also `strconv.Itoa` is how we convert between int to string.

AbhiPrasad (commented):
Please run `make backend-fmt`

We also need to write tests for this 1000% to confirm stuff.

Maybe even extracting the form parsing logic into it's own function and unit testing that.

restapi/stories.go
```diff
@@ -34,8 +34,65 @@ func init() {
...
+			sortString += sort[i] + " " + order[i]
+		}
+
+		if i != len(sort)-1 {
+			sortString += ", "
+		}
+	}
+
+	gormStories := api.database.Table("stories")
+	gormTags := api.database.Table("tags")
```
Shehryar21, line 114: We probably dont need to query Tags table in return stories endpoint. If you look at my PR, we now return tags as part of a new struct called Taggedstory. So in terms of sorting and querying by tag, we can use that. We can discuss this more if you are confused about this.

nimwijetunga: The thing is we don't want to manually pull all stories with tags then query them when they're in an array. We should call your function at the very end and instead use sql queries to do the filtering and sorting (easier logic wise and faster)

restapi/stories.go
```diff
@@ -34,8 +34,65 @@ func init() {
...
+
+	err = gormTags.Where("name IN ?", tags).Find(&storiesByTags).Error
+	if err != nil {
+		return rest.ErrInternal(api.logger, err)
+	}
+	fmt.Printf("storiesBytags: %v\n", storiesByTags)
```
nimwijetunga: remove this

restapi/stories.go
```diff
@@ -34,8 +34,65 @@ func init() {
...
+
+	if len(storiesByTags) != 0 {
+		fmt.Printf("in storiesByTags \n")
```
nimwijetunga: remove

restapi/stories.go
```diff
@@ -34,8 +34,65 @@ func init() {
...
+		}
+		gormStories = gormStories.Where("id IN ?", storyIDs)
+	}
+
+	if len(sortString) != 0 {
+		gormStories = gormStories.Order(sortString)
+		fmt.Printf("in sortString \n")
+	}
+
+	err = gormStories.Where("is_visible = ?", visibility).Find(&stories).Error
```
nimwijetunga, line 134: lets do this where query before doing the order (otherwise it might sometimes affect the sort). Ordering should always be last

restapi/stories.go
```diff
@@ -34,8 +34,65 @@ func init() {
...
 	if err != nil {
```
nimwijetunga, line 135: For the last line of this function can you add a TODO to use `ReturnTaggedStories(stories)` that will be added when Shehryar merges his pr

restapi/stories.go
```diff
@@ -34,8 +34,65 @@ func init() {
...
+		storyIDs := make([]string, len(storiesByTags))
+		for i := 0; i < len(storiesByTags); i++ {
+			storyIDs[i] = strconv.FormatUint(uint64(storiesByTags[i].StoryID), 10)
+		}
+		gormStories = gormStories.Where("id IN ?", storyIDs)
+	}
+
+	if len(sortString) != 0 {
+		gormStories = gormStories.Order(sortString)
+		fmt.Printf("in sortString \n")
```
nimwijetunga: remove

nimwijetunga (changes requested):
1st pass let me know if you have any questions

nimwijetunga (approved):
LGTM. Please test before merging

restapi/stories.go
```diff
@@ -77,8 +77,61 @@ func (api api) AddAuthorToStories(stories []models.Story) ([]models.Story, error
 
 func (api api) ReturnAllStories(w http.ResponseWriter, r *http.Request) render.Renderer {
 	var stories []models.Story
+	var storiesByTags []models.Tag
 
-	err := api.database.Where("is_visible = true").Find(&stories).Error
+	getVisibility := r.URL.Query()["visibility"]
```
Shehryar21, line 82: why not do r.FormValue["visibility"] to stay sort of consistent with how you are getting other query parameters. i think if it wont exist, it would default to ""

restapi/stories.go
```diff
@@ -77,8 +77,61 @@ func (api api) AddAuthorToStories(stories []models.Story) ([]models.Story, error
...
+		return rest.ErrInternal(api.logger, err)
+	}
+
+	if len(storiesByTags) != 0 {
+
+		storyIDs := make([]string, len(storiesByTags))
+		for i := 0; i < len(storiesByTags); i++ {
+			storyIDs[i] = strconv.FormatUint(uint64(storiesByTags[i].StoryID), 10)
+		}
+		gormStories = gormStories.Where("id IN ?", storyIDs)
```
Shehryar21, line 127: should we error check here?

restapi/stories.go
```diff
@@ -77,8 +77,61 @@ func (api api) AddAuthorToStories(stories []models.Story) ([]models.Story, error
...
+	}
+
+	if len(sortString) != 0 {
+		gormStories = gormStories.Order(sortString)
```
Shehryar21, line 131: same here, check for error?

Shehryar21 (approved):
LGTM! Great Work. Just added few small comments.

restapi/stories.go
DinuWije: There is no need to do an error check here because we aren't actually making a db call here (we are constructing the call). When the call is made below (via the .Find), we do an error check.

restapi/stories.go
DinuWije: same as above

restapi/stories.go
DinuWije: These are query params, not form values. The reason I used from values for the ones below is so I could store the results inside arrays (which is not needed for this)

PR #239, opened 2021-02-27
Edit story backend

### LINKED ISSUE #236: Edit Story - Backend
1. Fix getting the author along with the story
2. Make sure the story/{id} endpoint has all the fields displayed here: https://www.figma.com/file/uVqnweL2ApqvPHWzQYmZ1t/Dashboard?node-id=317%3A14
  -> getFormattedStory(story models.Story) taggedStory {
                   // contains all story fields and tags
      }
3. Update the story with all given fields
4. Delete the previous image and upload new one (deleting is a nice to have, we can just keep the old image in s3 for now)
5. Updating Authors
   -> update story table with author fn, ln, country
  -> If author doesn't exist create in the Authors table
  -> regardless of whether the author exists or not, update the bio
6. Updating tags
   -> delete all tags for a given story
   -> re insert if there is one or more tags

Shehryar21:
closes #236 

Delete image from s3 commented out for now because the way its done right now, if there are 2 images with same name, it deletes both of them which is not how it should be. We can take a look to fix later on.


restapi/stories.go
```diff
@@ -32,30 +32,148 @@ func init() {
 	rand.Seed(time.Now().UTC().UnixNano())
 }
 
+type TaggedStory struct {
+	models.Story
+	Tags []string `json:"tags"`
```
AbhiPrasad, line 38: plz update the typescript type in `types.tsx` to reflect the new story type being returned.

restapi/stories.go
```diff
@@ -27,35 +27,165 @@ import (
...
+			}
+			if err := api.database.Create(&author).Error; err != nil { //Adding author
+				return rest.ErrInternal(api.logger, errAuthor)
+			}
+		} else {
+			return rest.ErrInternal(api.logger, errAuthor)
+		}
+	}
+	author.Bio = r.FormValue("bio")
+	api.database.Save(&author)
```
AbhiPrasad: Can we wrap all the database operations (save, delete) in a single transaction -> https://gorm.io/docs/transactions.html.

Let's also do it at the end, I don't think we should be mixing db ops + setters for the story

restapi/stories.go
```diff
@@ -27,35 +27,165 @@ import (
...
+	story.Author = author
+	story.AuthorFirstName = r.FormValue("author_first_name")
+	story.AuthorLastName = r.FormValue("author_last_name")
+	story.AuthorCountry = r.FormValue("author_country")
+	story.ImageURL = imageURL
+	story.Title = r.FormValue("title")
+	story.Content = r.FormValue("content")
+	story.CurrentCity = r.FormValue("current_city")
```
AbhiPrasad: Can we just do a single story instantiation?

Like
```go
story := {
  Author: author,
  AuthorFirstName: r.FormValue("author_first_name")
  ...
}
```

Generally I wanna see this function organized like this:

form data |> processed form data |> story struct |> insert into db

That way we can easily extract one of these steps into a pure function for unit testing or to abstract so that other functions can use it.

AbhiPrasad (commented):
Quick comments on organizing the endpoint and how I want the function to look.

Man these are looking really solid good stuff.

restapi/stories.go
```diff
@@ -27,35 +27,165 @@ import (
...
+	err := api.database.Where("story_id=?", story.ID).Find(&tags).Error
+	if err != nil {
+		return TaggedStory{story, []string{}} //Test for this case
+	}
+	names := make([]string, len(tags))
+	for i, t := range tags {
+		names[i] = t.Name
+	}
+	TaggedStory := TaggedStory{story, names}
+	return TaggedStory
```
nimwijetunga: return `TaggedStory{story, names}` should be fine

restapi/stories.go
```diff
@@ -27,35 +27,165 @@ import (
...
+		return rest.ErrInvalidRequest(api.logger, "Error parsing year field", err)
+	}
+	story.Year = uint(year)
+	story.Summary = r.FormValue("summary")
+	city := story.CurrentCity
+	coordinates, err := api.locationFinder.GetCityCenter(city)
+	if err != nil {
+		return rest.ErrInvalidRequest(api.logger, "Story has an invalid current city", err)
+	}
+	story.Latitude = randomCoords(coordinates.Latitude)
```
nimwijetunga: Can you add a TODO here to change this to Megan's pin placement implementation once she has merged it in

restapi/stories.go
```diff
@@ -27,35 +27,165 @@ import (
...
+				return rest.ErrInvalidRequest(api.logger, "Unknown origin country", nil)
+			}
+			if err := api.database.Create(&author).Error; err != nil { //Adding author
+				return rest.ErrInternal(api.logger, errAuthor)
+			}
+		} else {
+			return rest.ErrInternal(api.logger, errAuthor)
+		}
+	}
+	author.Bio = r.FormValue("bio")
```
nimwijetunga: Since the author bio is an optional field you should make sure to check if the value exists. If it exists you can just set it (what you already have), if not then you will have to clear it

nimwijetunga (changes requested):
Looks great! Just a few small considerations

restapi/endpoints_test.go
```diff
@@ -145,7 +145,28 @@ func (suite *endpointTestSuite) TestReturnAllUniqueTags() {
 }
 
 func (suite *endpointTestSuite) TestGetAllStories() {
```
nimwijetunga, line 147: is it difficult to add a test for editing a story?

restapi/stories.go
```diff
@@ -40,22 +67,144 @@ func (api api) ReturnAllStories(w http.ResponseWriter, r *http.Request) render.R
 		return rest.ErrInternal(api.logger, err)
 	}
 
-	return rest.JSONStatusOK(stories)
+	for i, story := range stories {
```
nimwijetunga, line 70: Probably extract this into a function that returns an array of stories

restapi/endpoints_test.go
nimwijetunga, line 133: Is it difficult to add a test for editing a story?

restapi/stories.go
nimwijetunga, line 63: Probably extract this into a function that returns an array of stories

nimwijetunga (approved):
LGTM, just a few things to consider

restapi/endpoints_test.go
Shehryar21: It probably is. We are passing in formdata and also mocking uploading to s3 might be hard, but we can look into it. Would it make sense to look into testing after this gets merged in or right now?

nimwijetunga: Yeah lets get this merged first

nimwijetunga (approved):

AbhiPrasad (approved):

PR #240, opened 2021-03-02
feat: All stories skeleton 

hanlinc27:
ezgif com-gif-maker


ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,288 @@
...
+import { colors } from "../styles/colors";
+import { Story } from "../types";
+
+declare module "@material-ui/core/styles/withStyles" {
+  interface BaseCSSProperties {
+    flip?: boolean;
+  }
+}
+
+const styles = (theme: Theme) =>
```
hanlinc27, line 10: Could be changed to `styled-components `but I pretty much left the styling the same as the example.

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,288 @@
...
+    noClick: {
+      cursor: "initial",
+    },
+  });
+
+interface ColumnData {
+  dataKey: string;
+  label: string;
+  numeric?: boolean;
+  toggle?: true;
```
hanlinc27: I wanted to do something like `   {columns[columnIndex].toggle && (
          <Switch
            checked={checked}
            onChange={handleChange}
            color="primary"
          />
        )}`
but hooks should not be called conditionally.

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+  getRowClassName = ({ index }: Row) => {
+    const { classes, onRowClick } = this.props;
+
+    return clsx(classes.tableRow, classes.flexContainer, {
+      [classes.tableRowHover]: index !== -1 && onRowClick != null,
+    });
+  };
+
+  cellRenderer: TableCellRenderer = ({ cellData, columnIndex }) => {
```
hanlinc27: Can probably simplify the default cell header and column rendering logic from MUI

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,288 @@
...
+          </Table>
+        )}
+      </AutoSizer>
+    );
+  }
+}
+
+const VirtualizedTable = withStyles(styles)(MuiVirtualizedTable);
+
+interface TableData {
```
hanlinc27: Should I be defining a new interface for this? Or just loading in from `Story` types?

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,288 @@
...
+  country: string,
+  is_visible: boolean
+): TableData {
+  return { id, title, city, year, author, country, is_visible };
+}
+
+export const AllStories: React.FC = () => {
+  const { data: allStories, error } = useSWR<Story[]>("/api/stories");
+
+  if (allStories) {
```
hanlinc27, line 69: Need an **if statement** here, otherwise we get undefined at [0] errors.

ui/src/pages/AllStories.tsx
hanlinc27: ^ in cell rendering

ui/src/pages/AllStories.tsx
DinuWije: I think this is fine considering the demo you posted looks good and we are tight on time to finish the project. Correct me if I'm wrong, but we can still mix styled-components in with createStyles elements right?

ui/src/pages/AllStories.tsx
DinuWije: I may be misinterpreting your comment, but I think it's fine to call hooks conditionally as long as they aren't **instantiated** conditionally. @AbhiPrasad pls take a look at this too in case I'm very wrong lmao.

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,288 @@
...
+    const rows: TableData[] = [];
+    for (let i = 0; i < 33; i += 1) {
```
DinuWije: Is the 33 here in the for statement because we currently have 33 stories? Can we put a variable instead (ie. allStories.length())?

ui/src/types/index.tsx
```diff
@@ -12,7 +12,7 @@ export interface Story extends Model {
   title: string;
   content: string;
   current_city: string;
-  year: number;
+  year: string;
```
DinuWije: what was the reasoning behind switching this to a string?

ui/package.json
```diff
@@ -30,13 +30,14 @@
     "leaflet.markercluster": "^1.4.1",
     "material-ui-dropzone": "^3.5.0",
     "prop-types": "^15.7.2",
-    "react": ">= 16.8",
-    "react-dom": ">= 16.8",
+    "react": "^15.3.0 || ^16.0.0-alpha",
+    "react-dom": "^15.3.0 || ^16.0.0-alpha",
```
AbhiPrasad, line 34: Why the changes here? We need 16.8 for hooks.

ui/src/pages/AllStories.tsx
AbhiPrasad: Yup according to the rules of hooks you cannot do something like this:

```ts
if (blah) {
  const [state, setter] = React.useState();
}
```

To understand this, you can imagine each useState call being a representing an index in an state array.

So imagine a ReactElement class that represents a React component in the virtual dom.

```
ReactElement {
  state: [state1, state2, state3, ...],
  effects: [...],
  callbacks: [...],
  ref: ...,
  children: [ ... ]
}
```

The first time a component is created, this `state` array is instantiated based on how many times `React.useState()` was called. The `setState` setters from `React.useState()` then rely on the array index to update state. Therefore if you don't call a `React.useState()` because of conditional reasons on a re-render, state/state-setter might reference the wrong index of the state array.

ui/src/pages/AllStories.tsx
AbhiPrasad: Yeah let's just escape hatch for now, we can come back to this.

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,288 @@
...
+  Table,
+  TableCellRenderer,
+  TableHeaderProps,
+} from "react-virtualized";
+import useSWR from "swr";
+
+import { colors } from "../styles/colors";
+import { Story } from "../types";
+
+declare module "@material-ui/core/styles/withStyles" {
```
AbhiPrasad: let's move this into `global.d.ts` plz

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,288 @@
...
+interface MuiVirtualizedTableProps extends WithStyles<typeof styles> {
+  columns: ColumnData[];
+  headerHeight?: number;
+  onRowClick?: () => void;
+  rowCount: number;
+  rowGetter: (row: Row) => TableData;
+  rowHeight?: number;
+}
+
+class MuiVirtualizedTable extends React.PureComponent<
```
AbhiPrasad: Can we extract this into it's own component (`Table.tsx`) and make them all functional? We aren't using any of the class based lifecycles.

We can just turn the class function into standalone functions.

AbhiPrasad: it costs nothing to pass down additional function args instead of using `this.props`

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,288 @@
...
+    },
+    noClick: {
+      cursor: "initial",
+    },
+  });
+
+interface ColumnData {
+  dataKey: string;
+  label: string;
+  numeric?: boolean;
```
AbhiPrasad: we shouldn't mix undefined and boolean types. Let's enforce this to be a boolean and write a data transform if needed.

ui/src/pages/AllStories.tsx
AbhiPrasad: Also toggle should just be a boolean, we can handle taking care of it when we grab the data instead of typing the component weirdly.

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,288 @@
...
+function createData(
+  id: number,
+  title: string,
+  city: string,
+  year: string,
+  author: string,
+  country: string,
+  is_visible: boolean
+): TableData {
+  return { id, title, city, year, author, country, is_visible };
```
AbhiPrasad: Can we memoize based on id? Aka keep a running hashmap that stores id -> `TableData` object?

ui/src/pages/AllStories.tsx
AbhiPrasad: I think we should prefer just using the Story type.

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,288 @@
...
+}
+
+export const AllStories: React.FC = () => {
+  const { data: allStories, error } = useSWR<Story[]>("/api/stories");
+
+  if (allStories) {
+    const rows: TableData[] = [];
```
AbhiPrasad: prefer `array.map` here.

So something like:

```
const { data: allStories, error } = useSWR<Story[]>("/api/stories");

if (allStories === undefined) return <div>Loading all stories table..</div>;

const rows = allStories.map((story, i) => createData(i, story.title, ...));
```

ui/package.json
AbhiPrasad: We should just pin to `16.8` directly

AbhiPrasad (changes requested):
1st pass review

ui/src/types/index.tsx
hanlinc27: The endpoint on the backend looks like this: 
image
so I just updated the types.

Shehryar21: So i think its better to keep it int because when stories are returned, year is an int. It is just that in formdata everything is in string format, so even if you send it as an int, it would be received in the backend as a string.

hanlinc27: Changed it back 👍🏻

hanlinc27:
Closing this and opening a new PR with updated changes/different approach.

PR #248, opened 2021-03-06
feat: All Stories Page: sorting, switches

### LINKED ISSUE #227: All Stories Tab Skeleton - Frontend
Designs are here: [https://www.figma.com/file/uVqnweL2ApqvPHWzQYmZ1t/Dashboard?node-id=2%3A100](https://www.figma.com/file/uVqnweL2ApqvPHWzQYmZ1t/Dashboard?node-id=2%3A100)

- For filtering and sorting everything will be done using query parameters to the backend all stories endpoint. Look at the [backend ticket ](https://github.com/uwblueprint/shoe-project/issues/226) to get the query parameters you will need to call to sort/filter.
- For performance, use something like https://github.com/bvaughn/react-window to address possible memory/performance concerns (lets not render a billion dom nodes)

hanlinc27:
#### Description 
In this PR we create two components, `VirtualizedTable` and `AllStoriesTabs`, which we import into our page at **/stories** in `AllStories.tsx`. We handle all sorting of each column, as well as toggle state/checked state on the client side.

#### All Stories
ezgif com-gif-maker

#### Tabbed Header
image

#### Todo: 
* styling fixes and header + logo
* break down visible changes and pending map changes into two tickets

#### Bugs:
This PR also has a console bug that goes away on refresh 
image
`theme.spacing.unit` is either used in the `Checkbox` or `Table` MUI components. We are on v4, it fixes if we're willing to [revert back to v3](https://github.com/gregnb/mui-datatables/issues/676).


ui/package.json
```diff
@@ -23,14 +23,15 @@
...
+    "@material-ui/core": "^4.11.3",
     "@material-ui/icons": "^4.0.0",
     "@material-ui/lab": "^4.0.0-alpha.56",
     "leaflet": "^1.0.2",
     "leaflet.markercluster": "^1.4.1",
     "material-ui-dropzone": "^3.5.0",
+    "mui-virtualized-table": "^2.2.3",
     "prop-types": "^15.7.2",
-    "react": ">= 16.8",
+    "react": ">=15",
```
hanlinc27, line 34: Wiill pin back

ui/src/App.tsx
```diff
@@ -35,6 +35,9 @@ function App(): JSX.Element {
             <Route path="/upload">
               <Upload />
             </Route>
+            <Route path="/stories">
```
hanlinc27: Should this be under protected routes?

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,210 @@
...
+        story.year,
+        story.is_visible,
+        story.author_first_name,
+        story.author_last_name,
+        story.author_country
+      )
+    );
+  }
+
+  const [visibleState, setVisibleState] = React.useState([]);
```
hanlinc27, line 62: put into useReducer?

hanlinc27 (commented):
Another stab at all stories page

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const [tableData, setTableData] = useState(rows);
+  const [selectedRowIds, setSelectedRowIds] = useState([]);
+  const [order, setOrder] = useState("desc");
+  const [orderBy, setOrderBy] = useState("age");
+
+  const handleChange = (e, d) => {
+    if (e.target.checked) {
```
hanlinc27, line 69: We know what story the checkbox/toggle state belongs to based off of ID. We can also create a tabledata type but there isn't one currently.

ui/src/App.tsx
AbhiPrasad: Yup, I'll take care of it after you get it merged though.

ui/src/components/VirtualizedTable.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <AutoSizer>
+        {({ width }) => (
+          <MuiTable
+            data={data}
+            columns={columns}
+            orderBy={orderBy}
+            orderDirection={order}
+            includeHeaders
+            cellProps={(column, row) => {
+              if (data.indexOf(row) > 0 && data.indexOf(row) % 2 !== 0) {
```
AbhiPrasad, line 57: plz add comment explaining why you have this boolean condition

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,210 @@
...
+  id: number,
+  title: string,
+  current_city: string,
+  year: number,
+  is_visible: boolean,
+  author_first_name: string,
+  author_last_name: string,
+  author_country: string
+) {
+  const author_name = author_first_name + " " + author_last_name;
```
AbhiPrasad:
```suggestion
  const author_name = `${author_first_name} ${author_last_name}`;
```

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,210 @@
...
+    author_name,
+    author_country,
+    is_visible,
+  };
+}
+
+export const AllStories: React.FC = () => {
+  const { data: allStories, error } = useSWR<Story[]>("/api/stories");
+  let rows = [];
+  if (allStories) {
```
AbhiPrasad, line 46: If we do an early check for `allStories === undefined` (and do an early return), we can just do `const rows = allStories.map` and we don't need the mutable rows variable + do the boolean check for `allStories`

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,210 @@
...
+        story.author_country
+      )
+    );
+  }
+
+  const [visibleState, setVisibleState] = React.useState([]);
+  const [tableData, setTableData] = useState(rows);
+  const [selectedRowIds, setSelectedRowIds] = useState([]);
+  const [order, setOrder] = useState("desc");
+  const [orderBy, setOrderBy] = useState("id");
```
AbhiPrasad, line 66: From the figma designs it seems that you can't order based on the table, only using the filter dropdown. This means we can probably delete this stuff for now.

I think we are doing filtering/sorting through pinging the backend, not through updating client state. This means we have to just store the query parameters and update the hook accordingly.

hanlinc27: Hmm Pam told me earlier that we want to do alphabetical sorting (asc/desc) for each individual column via toggle. Should we do this via the frontend and leave filtering to the backend queries? Or have both be just backend calls.

AbhiPrasad: Yeah that makes sense to me, leave sorting to the frontend and filtering to the backed. Let's make sure we bring this up in the meeting today if thats what we decide to do.

ui/src/pages/AllStories.tsx
hanlinc27: That causes the hooks to be conditional I think?

hanlinc27: allStories always === undefined in the beginning, which gives us an early return and means nothing gets rendered on the page.

ui/src/App.tsx
DinuWije: Maybe have the path as "/all-stories" or "/allstories" (or something along those lines). This would be to avoid confusion with the /api/stories backend endpoint. If you think this isn't important though, it's fine as is.

DinuWije (approved):
Everything looks good to me so I'm going to go ahead and approve but I'd def recommend having someone with a more trained eye taking a second look as well.

Also, I see you trying to reach my record of 8000 lines in a commit with ur yarn lock. Nice try but not quite.

ui/src/App.tsx
hanlinc27: makes sense! 👍🏻

liu-jeffrey (approved):

ui/src/App.tsx
```diff
@@ -35,6 +35,9 @@ function App(): JSX.Element {
             <Route path="/upload">
               <Upload />
             </Route>
+            <Route path="/allstories">
```
nimwijetunga, line 38: i think we should do `/stories` instead, makes a bit more sense

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,250 @@
...
+  id: number,
+  title: string,
+  current_city: string,
+  year: number,
+  is_visible: boolean,
+  author_first_name: string,
+  author_last_name: string,
+  author_country: string
+) {
+  const author_name = `${author_first_name} ${author_last_name}`;
```
Shehryar21, line 25: pretty small nitpick lol, but why not have this in return statement directly, if its only used once

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,250 @@
...
+`;
+
+export const AllStories: React.FC = () => {
+  const { data: allStories, error } = useSWR<Story[]>("/api/stories");
+  let rows = [];
+
+  // if (allStories===undefined) return;
+  if (allStories) {
+    rows = allStories.map((story, i) =>
+      createData(
```
Shehryar21, line 48: actually why even have a createData function if adding first name and last name is all what we are doing

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,250 @@
...
+  const getComparator = (order, orderBy) => {
+    return order === "desc"
+      ? (a, b) => descendingComparator(a, b, orderBy)
+      : (a, b) => -descendingComparator(a, b, orderBy);
+  };
+  const descendingComparator = (a, b, orderBy) => {
+    if (b[orderBy] < a[orderBy]) {
+      return -1;
+    }
+    if (b[orderBy] > a[orderBy]) {
```
Shehryar21: why not else if?

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,250 @@
...
+                <div>
+                  <Checkbox
+                    checked={selectedRowIds.length > 0}
+                    indeterminate={indeterminate}
+                    onChange={handleCheckedAll}
+                  />
+                  ID
+                </div>
+              ),
+              cell: (d) => (
```
Shehryar21, line 171: whats d here lol?

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,250 @@
...
+    if (e.target.checked) {
+      setVisibleState((prevState) => [...prevState, d.id]);
+    } else {
+      setVisibleState((prevState) => prevState.filter((e) => e !== d.id));
+    }
+  };
+
+  const handleCheckedAll = () => {
+    setSelectedRowIds((prevState) =>
+      prevState.length === rows.length ? [] : rows.map((d) => d.id)
```
nimwijetunga, line 78: I'm not too familiar with reach but where is `d` defined?

ui/src/pages/AllStories.tsx
```diff
@@ -0,0 +1,250 @@
...
+    if (e.target.checked) {
+      setSelectedRowIds((prevState) => [...prevState, d.id]);
+    } else {
+      setSelectedRowIds((prevState) => prevState.filter((e) => e !== d.id));
+    }
+  };
+
+  const handleRequestSort = (property) => {
+    const isDesc = orderBy === property && order === "desc";
+    const newOrder = isDesc ? "asc" : "desc";
```
nimwijetunga: Is this flipping descending to ascending and vise-versa?

nimwijetunga (approved):
Great work!

ui/src/pages/AllStories.tsx
Shehryar21: I am probably wrong, but shouldnt it be `newOrder = isDesc ? "desc" : "asc" ;`

Shehryar21 (approved):

ui/src/pages/AllStories.tsx
hanlinc27: d is defined in the function (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)

ui/src/pages/AllStories.tsx
hanlinc27: yup!

ui/src/pages/AllStories.tsx
hanlinc27: I think it's cleaner this way?

ui/src/pages/AllStories.tsx
hanlinc27: You need to have a data transform function here to ingest the data from our hook. Think about it this way, we need to have an object for each row of the table. We can just use the type `Story`, but it's cleaner if we select only the column data types we need.

ui/src/pages/AllStories.tsx
hanlinc27: you can actually just do `cell: ()` but it gives a lint warning for anonymous function so I just defined a param there.

ui/src/pages/AllStories.tsx
hanlinc27: It's initially desc which is why it's switched, but I can change that! :)

PR #255, opened 2021-03-13
View story

aspensantos:
https://user-images.githubusercontent.com/35615166/111043854-f8009e80-8412-11eb-84d1-0a8c46b59a0b.mp4


hanlinc27 (approved):
It works super well for me! Small styling - when I go from Upload --> Preview, the heading text is shifted up? Going from AllStories --> Story Preview looks fine though. Screenshot below for reference.
image

ui/src/pages/Upload.tsx
```diff
@@ -148,6 +151,63 @@ export const Upload: React.FC = () => {
...
+        longitude: 0,
+        ID: 0,
+        CreatedAt: "",
+        UpdatedAt: "",
+        DeletedAt: {
+          Time: "",
+          Valid: true,
+        },
+      };
+      console.log(storyFromData);
```
nimwijetunga: remove please

ui/src/pages/Upload.tsx
```diff
@@ -1,4 +1,5 @@
 import {
```
nimwijetunga, line 1: One thing you'll have to consider is doing the same thing for our edit story page (once that's merged in). How flexible is this logic?

ui/src/pages/AllStories.tsx
```diff
@@ -8,6 +8,7 @@ import * as React from "react";
 import styled from "styled-components";
```
nimwijetunga, line 8: I'm not entirely sure what this is doing but do you link the all stories page to this view story page here or is it something else?

nimwijetunga: If not we'll also need to add this logic in a separate pr

nimwijetunga (approved):
Works for me!

ui/src/pages/AllStories.tsx
aspensantos: yes when u click an entry in the table it opens the view story page for that entry

ui/src/pages/Upload.tsx
aspensantos: Yeah for edit story we would do the same thing, I think the form is essentially the same it is just populated so we should be able to use the same logic to create the story object and pass that to the view story drawer

ui/src/pages/Upload.tsx
aspensantos: done

PR #256, opened 2021-03-13
chore(deps): yarn-lock dependencies 

hanlinc27:
To confirm this works please
delete `.cache` file in /ui, delete yarn-lock file
`yarn`
`yarn watch`
`make backend`
This tentatively seems to fix the https://github.com/mui-org/material-ui/issues/10554 and https://github.com/parcel-bundler/parcel/issues/2957 errors we were seeing? 
Is the right way to do this supposed to be pushing up the yarn-lock file? Regardless, if you run **yarn** with this new package.json it should work and generate a correct lock file too. 
image
image

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -26,6 +26,7 @@
     "@material-ui/core": "^4.11.3",
     "@material-ui/icons": "^4.0.0",
     "@material-ui/lab": "^4.0.0-alpha.56",
+    "@material-ui/styles": "^4.11.3",
     "leaflet": "^1.0.2",
     "leaflet.markercluster": "^1.4.1",
     "material-ui-dropzone": "^3.5.0",
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -35,7 +35,7 @@ function App(): JSX.Element {
             <Route path="/upload">
               <Upload />
             </Route>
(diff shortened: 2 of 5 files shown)


nimwijetunga:
Even after deleting and recloning the repo, I'm still getting a blank page on `/stories`

liu-jeffrey:
> Even after deleting and recloning the repo, I'm still getting a blank page on `/stories`

Try `http://localhost:8900/allstories` !!! 

Also how the heck do i reply to a comment?? does it just stick a new comment like this on github?? that's kinda icky. 

Anyways, everything works for me! Still gotta do that janky thing where you make a change in AllStories.tsx and then wait for the yarn watch hot-patch. I wonder how we can fix that? Does calling yarn build help? Maybe @AbhiPrasad has an idea :D

liu-jeffrey (approved):

AbhiPrasad (approved):
Hmm the yarn watch errors are very mysterious - it's definitely just parcel shenanigans. Not sure what advice I got for you @liu-jeffrey 

As someone constantly in js land FOMO i kinda wanna try using esbuild https://github.com/evanw/esbuild (it's extreme speed) - I would actually recommend all new projects to use it!!

PR #257, opened 2021-03-14
Fixed Image deletion s3

### LINKED ISSUE #246: Figure out S3 Image Deletion
Currently, all images with same name gets deleted. Look into how to solve this issue.

Shehryar21:
closes #246.

ToDo: For the existing stories.json data, it would be better to add versionId of the image in the imageURL field. The code works without this change as well, but it would be more robust if the versionID of the seeded stories is added in imageURL. This change in data could be done in another PR.


restapi/stories.go
```diff
@@ -27,7 +27,9 @@ import (
 const youtubeRegex = `(?:youtube\.com\/(?:[^\/]+\/.+\/|(?:v|e(?:mbed)?)\/|.*[?&]v=)|youtu\.be\/)([^"&?\/\s]{11})`
 const youtubeEmbedURL = "https://www.youtube.com/embed/"
 
-//const s3KeyNameRegex = `^https://shoeproject.s3.us-west-000.backblazeb2.com/(.*)`
+const s3KeyNameRegex = `^https:\/\/shoeproject.s3.us-west-000.backblazeb2.com\/(.*)\?versionId`
```
AbhiPrasad: Can we just call `regexp.MustCompile` and the it'll break on compile globally

restapi/stories.go
```diff
@@ -117,24 +119,26 @@ func (api api) EditStoryByID(w http.ResponseWriter, r *http.Request) render.Rend
...
 
 		if err != nil {
 			return rest.ErrInvalidRequest(api.logger, "ImageUrl of existing image is not valid", err)
 		}
 
-		msg, err := api.DeleteImageInS3(imageKey) //delete existing image
+		versionId, err := convertImageURLToVersionId(story.ImageURL)
+
+		if err != nil {
+			return rest.ErrInvalidRequest(api.logger, "ImageUrl of existing image is not valid", err)
```
nimwijetunga: lets change the error message here since its the same as above

restapi/stories.go
```diff
@@ -232,6 +235,22 @@ func (api api) DeleteStoryByID(w http.ResponseWriter, r *http.Request) render.Re
 		return rest.ErrInternal(api.logger, err)
 	}
 
+	imageKey, err := convertImageURLToKeyName(story.ImageURL)
```
nimwijetunga, line 238: Are we confident that two stories wont share the same image? I remember you mention that bacblaze uses the same image if the filename is the same?

nimwijetunga: If we're not sure I would suggest we add a guard to validate that story.ImageURL is unique (i.e only appears in one story) before deleting

restapi/stories.go
```diff
@@ -332,12 +351,11 @@ func (api api) uploadImageTos3(file multipart.File, size int64, name string) (st
 	if err != nil {
 		return "", fmt.Errorf("Could not upload to S3")
 	}
-	return fmt.Sprintf("https://%s.s3.us-west-000.backblazeb2.com/%s", os.Getenv("BUCKET_NAME"), name), nil
+	return fmt.Sprintf("https://%s.s3.us-west-000.backblazeb2.com/%s?versionId=%s", os.Getenv("BUCKET_NAME"), name, *res.VersionId), nil
```
nimwijetunga, line 354: remove this please

restapi/stories.go
```diff
@@ -346,29 +364,38 @@ func (api api) DeleteImageInS3(name string) (string, error) {
...
+		}
+	}
+	return match[0][1], nil
+}
+
+func convertImageURLToVersionId(originalURL string) (string, error) {
+	match := s3VersionIdRegex.FindAllStringSubmatch(originalURL, 2)
 	if len(match) == 0 {
-		return "", fmt.Errorf("Invalid Image Link")
+		return "", nil
```
nimwijetunga, line 392: whats the point of returning an error in this function if its always nil. I think in this case we should return an `fmt.Errorf`

nimwijetunga (changes requested):
Looks good overall, just a few things to consider

restapi/stories.go
Shehryar21: Remove what? VersionId is needed here.

nimwijetunga: oh my bad i thought you were logging something here nvm

nimwijetunga (approved):
LGTM

PR #263, opened 2021-03-17
feat: Countries - Backend

### LINKED ISSUE #260: Countries - Backend
Model in the backend:

- only column is country

Endpoints:

- seed the countries database with all valid countries (make sure only a few countries)

- POST /countries - adds country to table, 

Payload: [{
"country_name": "name"
},
...]
- GET /countries -  get all countries in DB 

select country: canada, us, .., random_country that doesn't legally exist

Shehryar21:
closes #260 

Includes seeding logic, get all, and post country endpoint

sample request to add countries:

image

### DIFF
--- a/restapi/authors.go
+++ b/restapi/authors.go
@@ -36,7 +36,7 @@ func (api api) CreateAuthors(w http.ResponseWriter, r *http.Request) render.Rend
 	return rest.MsgStatusOK("Authors added successfully")
 }
 
-func (api api) ReturnAllCountries(w http.ResponseWriter, r *http.Request) render.Renderer {
+func (api api) ReturnAllUsedCountries(w http.ResponseWriter, r *http.Request) render.Renderer {
 	var authors []models.Author
 	countries := make(map[string]bool)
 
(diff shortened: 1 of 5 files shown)


restapi/authors.go
```diff
@@ -36,7 +36,7 @@ func (api api) CreateAuthors(w http.ResponseWriter, r *http.Request) render.Rend
 	return rest.MsgStatusOK("Authors added successfully")
 }
 
-func (api api) ReturnAllCountries(w http.ResponseWriter, r *http.Request) render.Renderer {
+func (api api) ReturnAllUsedCountries(w http.ResponseWriter, r *http.Request) render.Renderer {
```
nimwijetunga: This name is kind of misleading, lets use `ReturnAuthorOriginCountries` or something along those lines

restapi/countries.go
```diff
@@ -0,0 +1,39 @@
...
+		return rest.ErrInternal(api.logger, err)
+	}
+	names := make([]string, len(countries))
+	for i, t := range countries {
+		names[i] = t.Name
+	}
+	return rest.JSONStatusOK(names)
+}
+
+func (api api) AddCountries(w http.ResponseWriter, r *http.Request) render.Renderer {
```
nimwijetunga, line 40: Can you document a sample payload here so frontend can use it as a reference when they call this endpoint?

nimwijetunga (approved):
LGTM, please make sure you test that these endpoints work as expected before merging

restapi/countries.go
Shehryar21: like in the code as a comment?

PR #264, opened 2021-03-17
fix: All Stories table errors

### LINKED ISSUE #261: [Bug] AllStories - MaterialUI Dependency
image

Hot reload not working properly.

hanlinc27:
@liu-jeffrey This should fix #261 even though turns out it wasn't even a MUI error!

What was breaking: 
- the `mui-virtualized-table` package had to be updated to a pre-release (https://github.com/techniq/mui-virtualized-table/releases/tag/v3.0.0-6) so that the theme usage and `componentWillReceiveProps` deprecation gets fixed (I was tempted to also upgrade our `material-ui/core` to the new v5.0.0-alpha.27 but I think we're good on v4 right now.
- however, this wasn't the reason why we needed two builds --> `stableSort = (array, comparator)` was returning an empty array on the first time it was sent in as a prop to `VirtualizedTable`. Maybe because it was trying to sort empty table data that hadn't finished fetching? Either way, adding a conditional check for that fixed it right away. :)

Please pull and test locally to confirm it works. 

Closes #261


aspensantos (approved):

liu-jeffrey (approved):
Solid 10/10

You might have to empty cache and hard reload your browser to see these changes. To do this, click CTRL + SHIFT + I, then right click on the refresh button in Chrome, and select "Empty Cache and Hard Reload".

PR #266, opened 2021-03-17
Client Based Auth - Backend Modifications

nimwijetunga:
- Given a valid token from the client the backend will validate the user and create an entry in the DB


restapi/auth.go
```diff
@@ -33,47 +34,29 @@ func generateStateOauthCookie(w http.ResponseWriter) (string, error) {
...
+		Email: googleClaims.Email,
+		Hd: googleClaims.Hd,
 	}
 
-	// redirect if invalid user
+	// create user in database if valid
 	err = api.database.FirstOrCreate(&user, models.User{Email: user.Email}).Error
 	if err != nil {
-		http.Redirect(w, r, "/unauthorized", http.StatusTemporaryRedirect)
+		return rest.ErrInternal(api.logger, err)
```
nimwijetunga: I still generate the JWT token. I think we should keep the backend JWT token because:

1. The backend truly identifies when a valid user has logged in (i.e shoeproject or blueprint) so we shouldn't really be using auth tokens immediately provided by google
2. It generally doesn't make sense for a client to generate their own JWT token

nimwijetunga:
I'll fix lint issues at the end

restapi/auth.go
nimwijetunga: Right now its in the cookie but i wouldn't be opposed to just returning it and letting the frontend store it however they want

s278gupt: iirc we chose to store it in a cookie so that fe wouldn't need to worry about storage and retrieving & sending the token as part of the request (i.e. setting it in the req header every time)

restapi/auth.go
```diff
@@ -33,47 +34,29 @@ func generateStateOauthCookie(w http.ResponseWriter) (string, error) {
...
-func (api api) AuthCallback(w http.ResponseWriter, r *http.Request) render.Renderer {
-	expectedState, _ := r.Cookie("oauthstate")
-	if expectedState.Value != r.FormValue("state") {
-		return rest.ErrUnauthorized("Invalid oauth state")
-	}
+func (api api) Login(w http.ResponseWriter, r *http.Request) render.Renderer {
 
-	// exchange received authorization code for an access token
-	token, err := api.oauthconfig.Exchange(context.Background(), r.FormValue("code"))
+	payload, err := idtoken.Validate(oauth2.NoContext, r.Header.Get("Authorization"), config.GetGoogleClientId())
```
s278gupt: what's `idtoken`?

nimwijetunga: So the frontend will esentially handle the actual google sign in and when its successful google will return a token. This token is sent to the backend so we can verify and add the user to our db. `idtoken` is a library that is a wrapper around the google api that allows us to decode the token the frontend sends us

AbhiPrasad (approved):
Let's get this merged in?

PR #267, opened 2021-03-20
feat: Visible Stories and Pending Changes Tables

### LINKED ISSUE #251: All Stories Visible Changes - Frontend
https://www.figma.com/file/uVqnweL2ApqvPHWzQYmZ1t/Dashboard?node-id=551%3A2552 

Make table which shows visible changes.

### LINKED ISSUE #252: All Stories Pending Map Changes - Frontend
https://www.figma.com/file/uVqnweL2ApqvPHWzQYmZ1t/Dashboard?node-id=441%3A1144

Make table which shows pending map changes.

hanlinc27:
### PR Description 

- Visible Stories Table (Initial state is populated with all stories where `is_visible = true`. Toggling off the switches for visibility in the All Stories page will remove it from the table, same as from the Visible Stories page) 
- Pending Changes Table (this reflects changes either **-** or **+** to a story's visibility from the original backend state)
- Added styles to the table, switches, checkbox to reflect Figma

Note: `  mutate(
      "/api/stories",
      (prevStories: Story[]) => {
        return prevStories.map((currStory) =>
          currStory.ID === story.ID
            ? { ...currStory, is_visible: !currStory.is_visible }
            : currStory
        );
      },
      false
    );` 
is used for revalidating the data from `api/stories` to trigger our `useEffect` since our switches depend on `checked={story.is_visible}` 
https://swr.vercel.app/docs/mutation
### Demo

https://user-images.githubusercontent.com/19617248/112792957-67d97080-9032-11eb-82e3-5cd0c0793704.mp4

Closes #251
Closes #252


ui/src/pages/AllStories.tsx
```diff
@@ -10,8 +11,42 @@ import useSWR from "swr";
...
+  background-color: ${colors.primaryLight6};
+`;
+
+const useStyles = makeStyles({
+  root: {
+    background: colors.primaryLight6,
+    fontFamily: "Poppins",
+    fontStyle: "normal",
+    fontWeight: 500,
+    fontSize: "16px",
```
hanlinc27, line 39:
```suggestion
    fontSize: fontSize.subtitle
```

ui/src/pages/AllStories.tsx
```diff
@@ -60,17 +90,21 @@ export const AllStories: React.FC = () => {
 
   const [tabValue, setTabValue] = React.useState(0);
   const [visibleState, setVisibleState] = React.useState([]);
```
hanlinc27, line 92: Maybe more descriptive name?

ui/src/styles/typography.ts
```diff
@@ -311,3 +311,14 @@ export const UploadLabelsText = styled.div`
...
   color: ${colors.black};
 `;
+
+export const StyledAllStoriesHeader = styled.div`
+  font-family: Canela;
+  font-style: normal;
+  font-weight: normal;
+  font-size: ${fontSize.h1Text};
+  line-height: 37px;
+  padding-bottom: 24px;
```
hanlinc27, line 321:
```suggestion
  padding: 80px 0 24px 0;
```

hanlinc27 (commented):
Please let me know if anything is unclear/if it's not behaving locally. :)

hanlinc27:
Nvm, please don't review yet

hanlinc27:
Nvm, don't review yet

nimwijetunga:
CSS looks a bit weird for me when i run this locally: 
image

nimwijetunga:
Strange on my Mac its fine but on Linux its slightly messed up. Maybe if it works well on windows we can just chalk it up to an OS issue, noone really uses Linux anyways

hanlinc27:
> Strange on my Mac its fine but on Linux its slightly messed up. Maybe if it works well on windows we can just chalk it up to an OS issue, noone really uses Linux anyways

Was probably a not up to date version - the screenshot you sent was the styling from a few commits ago.

ui/src/components/VirtualizedTable.tsx
```diff
@@ -30,6 +36,7 @@ const useVirtualizedTableStyles = makeStyles({
 
 interface VirtualizedTableProps {
   data: [];
+  // eslint-disable-next-line
```
AbhiPrasad:
```suggestion
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```

AbhiPrasad: Also can we type plz?

```ts
type TableColumn = {
  name: string;
  width: number;
  onHeaderClick: () => void;
  header: JSX.Element;
  cell: () => JSX.Element;
}
```

ui/src/pages/AllStories.tsx
```diff
@@ -34,56 +83,86 @@ function createData(
...
-
-  const [tabValue, setTabValue] = React.useState(0);
-  const [visibleState, setVisibleState] = React.useState([]);
-  const [tableData, setTableData] = useState(rows);
-
+  const [tabValue, setTabValue] = useState(0);
+  const [visibleState, setVisibleState] = useState([]);
+  const [visibleTableState, setVisibleTableState] = useState([]);
+  const [tableData, setTableData] = useState([]);
+  const [changedVisibility, setChangedVisibility] = useState([]);
```
AbhiPrasad: Kk when we have a ton of `useState` let's refactor this into a big `useReducer`.

ui/src/pages/AllStories.tsx
```diff
@@ -34,56 +83,86 @@ function createData(
...
+          story.is_visible,
+          story.author_first_name,
+          story.author_last_name,
+          story.author_country
+        )
+      );
+      //Initialize state after allstories are mapped
+      setVisibleState(rows);
+      setVisibleTableState(rows);
+      setTableData(rows);
```
AbhiPrasad: Instead of calling 3 `setState` calls, we can do a single dispatch if we do `useReducer()`

ui/src/pages/AllStories.tsx
```diff
@@ -34,56 +83,86 @@ function createData(
...
     if (e.target.checked) {
-      setVisibleState((prevState) => [...prevState, d.id]);
+      setVisibleState((prevStories) => [...prevStories, story.id]);
+      setVisibleTableState([...visibleTableState, story]);
+    } else {
+      setVisibleState((prevStories) =>
+        prevStories.filter((e) => e !== story.id)
+      );
+      setVisibleTableState((visibleTableState) =>
+        visibleTableState.filter((e) => e.id !== story.id)
```
AbhiPrasad: Ya this complicated state logic can just be a single action being dispatched with an id if we use `useReducer()`

AbhiPrasad (changes requested):
First pass. Let's refactor the multiple `useState()` calls into a `useReducer()`

ui/src/pages/AllStories.tsx
```diff
@@ -1,17 +1,66 @@
...
+  },
+  indicator: {
+    backgroundColor: colors.primaryDark1,
+  },
+  checkbox: {
+    "&$checked": {
+      color: colors.primaryDark1,
+    },
+  },
+  checked: {},
```
aspensantos: remove this please :)

ui/src/pages/AllStories.tsx
```diff
@@ -1,17 +1,66 @@
...
+    lineHeight: "150%",
+    color: colors.primaryDark1,
+    paddingLeft: "64px",
+    boxShadow: "none",
+  },
+  indicator: {
+    backgroundColor: colors.primaryDark1,
+  },
+  checkbox: {
+    "&$checked": {
```
aspensantos: I have always seen people use "&:checked" but maybe im wrong

ui/src/pages/AllStories.tsx
```diff
@@ -34,56 +83,86 @@ function createData(
...
   const [selectedRowIds, setSelectedRowIds] = useState([]);
   const [order, setOrder] = useState("desc");
   const [orderBy, setOrderBy] = useState("id");
 
-  const handleChange = (e, d) => {
+  let rows = [];
+  useEffect(() => {
+    if (allStories) {
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+      rows = allStories.map((story) =>
```
aspensantos: pretty sure we can do this processing in a fetcher passed to swr This will avoid an extra rerender caused by this useeffect
```
const fetcher = url => axios.get(url).then(res => res.data.map((story) => createData(
          story.ID,
          story.title,
          story.current_city,
          story.year,
          story.is_visible,
          story.author_first_name,
          story.author_last_name,
          story.author_country
        )
      )))

**Review by aspensantos — APPROVED** [review]

File: ui/src/pages/AllStories.tsx [hunk 11]
```diff
@@ -21,111 +85,247 @@ function createData(
...
+    image_url,
+    video_url,
+    content,
   };
 }
 
-const StyledAppBar = styled(AppBar)`
-  //styling
-`;
+interface State {
```
AbhiPrasad: kk let's extract this the reducer/action/state logic into another file plz. A folder structure like this?

```
pages
  ...
  AllStories -> folder
    index.tsx -> has the component
    reducer.ts -> has the reducer + state, action types
  ...
```

AbhiPrasad: You can also have a `types.tsx` file to protect against potenial circular dependencies.

ui/src/pages/AllStories.tsx
```diff
@@ -21,111 +85,247 @@ function createData(
...
-        story.author_first_name,
-        story.author_last_name,
-        story.author_country
-      )
-    );
+const initialState: State = {
+  tabValue: 0,
+  visibleState: [] as StoryView[],
+  visibleTableState: [] as StoryView[],
+  tableData: [] as StoryView[],
```
AbhiPrasad: We don't need to cast the types if we already set the record type above with `initialState: State`

ui/src/pages/AllStories.tsx
```diff
@@ -21,111 +85,247 @@ function createData(
...
+      return {
+        ...state,
+        tabValue: action.id,
+      };
+    }
+    case "HANDLE_SWITCH_CHANGE": {
+      const changedVisibilityFoundID = (elem) => elem.ID === action.story.ID;
+      const changedVisibilityContainsID = state.changedVisibility.some(
+        changedVisibilityFoundID
+      );
```
AbhiPrasad: Prefer: 
```suggestion
      const changedVisibilityContainsID = state.changedVisibility.some(
        (e) => e.ID === action.story.ID
      );
```

ui/src/pages/AllStories.tsx
```diff
@@ -21,111 +85,247 @@ function createData(
...
+      };
+    }
   }
+}
 
-  const [tabValue, setTabValue] = React.useState(0);
-  const [visibleState, setVisibleState] = React.useState([]);
-  const [tableData, setTableData] = useState(rows);
+export const AllStories: React.FC = () => {
+  const [state, dispatch] = useReducer(reducer, initialState);
```
AbhiPrasad: damn `useReducer` looking clean as hell.

ui/src/pages/AllStories.tsx
```diff
@@ -21,111 +85,247 @@ function createData(
...
+    const order = isDesc ? "desc" : "asc";
+    dispatch({ type: "SET_ORDERING", order, orderBy: property });
 
-    onChangeTableSort(newOrder, property);
+    onChangeTableSort(order, property);
   };
   const onChangeTableSort = (order, property) => {
-    const newData = stableSort(tableData, getComparator(order, property));
-    setTableData(newData);
+    const data = stableSort(state.tableData, getComparator(order, property));
```
AbhiPrasad: Can we move the `stableSort` function out of the component? We can put it into a `helpers.ts` file.

ui/src/pages/AllStories.tsx
```diff
@@ -141,60 +341,273 @@ export const AllStories: React.FC = () => {
...
+    state.selectedRowIds.length !== allStories.length;
   if (error) return <div>Error returning stories data!</div>;
   if (!allStories) return <div>Loading all stories table..</div>;
 
   const handleTabChange = (
     event: React.ChangeEvent<Record<string, unknown>>,
     newValue: number
   ) => {
-    setTabValue(newValue);
+    // setTabValue(newValue);
```
AbhiPrasad: Can we get 🔥 ?

ui/src/pages/AllStories.tsx
```diff
@@ -1,18 +1,82 @@
...
+  },
+  checkbox: {
+    "&$checked": {
+      color: colors.primaryDark1,
+    },
+  },
+  checked: {},
+});
+
+export interface StoryView {
```
AbhiPrasad: Can we make `StoryView` a intersection type? so something like:

```ts
type StoryView = Story & {
  // whatever fields have changed or been added
}
```

ui/src/pages/AllStories.tsx
```diff
@@ -21,111 +85,247 @@ function createData(
...
-  const [order, setOrder] = useState("desc");
-  const [orderBy, setOrderBy] = useState("id");
+  const fetchStories = (url) =>
+    fetch(url)
+      .then((res) => {
+        return res.json();
+      })
+      .then((response) =>
+        response.payload.map((story) => {
+          return createData(
```
AbhiPrasad: let's just pass in a story and then make `createData()` destructure:

```ts
return createData(story)

...

function createData({ ID, title, current_city ...}) {
 // ...
}

```

https://hacks.mozilla.org/2015/05/es6-in-depth-destructuring/ -> cntrl-F to object destructuring

ui/src/pages/AllStories.tsx
```diff
@@ -21,111 +85,247 @@ function createData(
...
+          );
+        })
+      );
 
-  const [clickedStory, setClickedStory] = useState<Story | undefined>(
+  const { data: allStories, error } = useSWR<StoryView[] | undefined>(
+    "/api/stories",
+    fetchStories
+  );
+  const [clickedStory, setClickedStory] = useState<StoryView | undefined>(
```
AbhiPrasad: Let's move this into `useReducer`?

ui/src/pages/AllStories.tsx
```diff
@@ -21,111 +85,247 @@ function createData(
...
     undefined
   );
+  const classes = useStyles();
+
+  useEffect(() => {
+    dispatch({ type: "INITIALIZE_AFTER_API", rows: allStories });
```
AbhiPrasad: we should only dispatch if `allStories` is defined. Add a TODO to handle error case (if allStories stays undefined or if we get an error.

ui/src/pages/AllStories.tsx
```diff
@@ -21,111 +85,247 @@ function createData(
...
-  const [tableData, setTableData] = useState(rows);
+export const AllStories: React.FC = () => {
+  const [state, dispatch] = useReducer(reducer, initialState);
 
-  const [selectedRowIds, setSelectedRowIds] = useState([]);
-  const [order, setOrder] = useState("desc");
-  const [orderBy, setOrderBy] = useState("id");
+  const fetchStories = (url) =>
+    fetch(url)
+      .then((res) => {
```
AbhiPrasad: What happens if there is an error? What should the `.catch()` logic here be?

ui/src/pages/AllStories.tsx
AbhiPrasad: Also great time to write unit tests for the reducer (check out what I did here: https://github.com/uwblueprint/shoe-project/pull/280/files#diff-013eb0dfde8f991d00c2f99a2ab0539ab325daba7b2905cfb69f108c5b255979R7)

AbhiPrasad (changes requested):
First pass review, `useReducer` looks really good, just some changes around moving things around.

AbhiPrasad:
Man those typed interfaces 😍

ui/src/pages/AllStories.tsx
hanlinc27: hmm I think it's unnecessary just for the one function? I think it's better to just keep it with the other `tableSort` functions

ui/src/pages/AllStories.tsx
hanlinc27: I'm having issues with stopping the event propagation from the `MouseClick` for viewing stories into the `SyntheticEvent` for the Switch toggling and Checkbox so not going to move it into a reducer until that works properly.

ui/src/pages/AllStories.tsx
hanlinc27: the SWR hook is using the fetcher for getting `allStories` which means it has state handling for the data fetched + the error state, right? Do we need another error handling in the fetcher?

ui/src/pages/AllStories.tsx
hanlinc27: Isn't intersection types for adding together different pre-existing types? in that case `StoryView` would end up having everything on `Story` +  the member `author_name` which we don't want since this would mean returning extra longitude, latitude, etc in `createData`?

ui/src/pages/AllStories.tsx
hanlinc27: ooh yes that cleans it up nicely

hanlinc27: thanks for the example, added some tests ✨  - wasn't too sure how to mock the ones requiring event listeners 
image

ui/src/pages/AllStories.tsx
AbhiPrasad: Ah true. Can we make use of the `Omit` or `Pick` type then? https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys

ui/src/__tests__/util/allstories.test.ts
```diff
@@ -0,0 +1,87 @@
+import {
```
AbhiPrasad: This should follow the same path as the pages.

```
ui/src/__tests__/pages/AllStories.test.ts
```

ui/src/__tests__/util/allstories.test.ts
```diff
@@ -0,0 +1,87 @@
...
+];
+
+describe("allstories table", () => {
+  describe("reducer", () => {
+    it("should set the correct tab value", () => {
+      const action: Action = { type: "SWITCH_TAB", id: INIT_STATE.tabValue };
+      expect(INIT_STATE.tabValue).toBe(0);
+      const newState = allStoriesReducer(INIT_STATE, action);
+      expect(newState.id !== 0).toBe(true);
+    });
```
AbhiPrasad: I'm confused by this test. Shouldn't it do something like?

```suggestion
    it("should set the correct tab value", () => {
      const newId = 3;
      const action: Action = { type: "SWITCH_TAB", id: newId };
      expect(INIT_STATE.tabValue).toBe(0);
      const newState = allStoriesReducer(INIT_STATE, action);
      expect(newState.tabValue).toBe(newId);
    });
```

ui/src/pages/AllStories/index.tsx
```diff
@@ -0,0 +1,542 @@
...
+
+  const fetchStories = (url) =>
+    fetch(url)
+      .then((res) => {
+        return res.json();
+      })
+      .then((response) =>
+        response.payload.map((story) => {
+          return createData(story);
+        })
```
AbhiPrasad:
```suggestion
        response.payload.map(createData)
```

ui/src/pages/AllStories/index.tsx
```diff
@@ -0,0 +1,542 @@
...
+  current_city,
+  year,
+  author_first_name,
+  author_last_name,
+  author_country,
+  is_visible,
+  image_url,
+  video_url,
+  content,
+}): StoryView {
```
AbhiPrasad:
```suggestion
}: Story): StoryView {
```

ui/src/pages/AllStories/reducer.ts
```diff
@@ -0,0 +1,123 @@
...
+  tableData: StoryView[];
+  changedVisibility: StoryView[];
+  selectedRowIds: number[];
+  order: "asc" | "desc";
+  orderBy: string;
+}
+
+export type Action =
+  | { type: "SWITCH_TAB"; id: number }
+  | { type: "HANDLE_SWITCH_CHANGE"; e: React.ChangeEvent; story: StoryView }
```
AbhiPrasad, line 16: To write a test for the react change event we can just mock it out, and then cast it to make typescript happy.

```ts
const changeEvent = {
    target: {
      checked: true, // replace with false depending on what condition you wanna test
    },
  } as React.ChangeEvent;
```

AbhiPrasad (commented):
Some more comments, hold on im not done reviewing tho so I'll come back to this.

ui/src/__tests__/util/allstories.test.ts
hanlinc27: yeah you're right that test made no sense. Maybe writing these at 2am is not the way to go.

ui/src/pages/AllStories.tsx
hanlinc27: Viewing stories does work but we're going to hold off on putting it into reducer b/c we might change the workflow to view a story --> click on name vs table

ui/src/pages/AllStories/index.tsx
```diff
@@ -0,0 +1,556 @@
...
+  const [clickedStory, setClickedStory] = useState<StoryView | undefined>(
+    undefined
+  );
+  const classes = useStyles();
+
+  const fetchStories = (url) =>
+    fetch(url)
+      .then((res) => {
+        return res.json();
+      })
```
AbhiPrasad:
```suggestion
      .then((res) => res.json())
```

ui/src/pages/AllStories/index.tsx
```diff
@@ -0,0 +1,556 @@
...
+  };
+
+  const handleRequestSort = (property: string) => {
+    const isDesc = state.orderBy === property && state.order === "asc";
+    const order = isDesc ? "desc" : "asc";
+    dispatch({ type: "SET_ORDERING", order, orderBy: property });
+
+    onChangeTableSort(order, property);
+  };
+  const onChangeTableSort = (order, property) => {
```
AbhiPrasad: Let's declare this before `handleRequestSort`

ui/src/pages/AllStories/index.tsx
```diff
@@ -0,0 +1,556 @@
...
+      return 1;
+    }
+    return 0;
+  };
+
+  const indeterminate =
+    state.selectedRowIds.length > 0 &&
+    state.selectedRowIds.length !== allStories.length;
+  if (error) return <div>Error returning stories data!</div>;
+  if (!allStories) return <div>Loading all stories table..</div>;
```
AbhiPrasad: Move this below `handleTabChange` functions

AbhiPrasad (approved):
Still think there some small things to clean up, but mostly just nits on myside. Since theres more work to be done here I think its fine to ship this for now.

Also, and this will require some more refactoring so we can do this in a future PR, but we should have specific routes for each table view (like pending should be `/admin#pending`)

We can use a https://reactrouter.com/web/api/HashRouter to accomplish this.

PR #274, opened 2021-03-27
Publish Stories - Backend

### LINKED ISSUE #273: Publish Stories - Backend
Create an endpoint that updates the stories visibility. Takes in a list of stories, and updates database accordingly.

Shehryar21:
closes #273 

The request payload is an array of stories similar to the array of stories returned be get all stories endpoint.

### DIFF
--- a/restapi/api.go
+++ b/restapi/api.go
@@ -63,6 +63,7 @@ func Router(db *gorm.DB, locationFinder location.LocationFinder) (http.Handler, 
 		rest.GetHandler(r, "/story/{storyID}", api.ReturnStoryByID)
 		rest.DeleteHandler(r, "/story/{storyID}", api.DeleteStoryByID)
 		rest.PutHandler(r, "/story/{storyID}", api.EditStoryByID)
+		rest.PutHandler(r, "/publish", api.PublishStories)
 		rest.GetHandler(r, "/authors/origin_countries", api.ReturnAuthorOriginCountries)
 		rest.GetHandler(r, "/tags", api.ReturnAllUniqueTags)
 		rest.GetHandler(r, "/login", api.Login)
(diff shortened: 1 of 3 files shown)


restapi/api.go
```diff
@@ -63,6 +63,7 @@ func Router(db *gorm.DB, locationFinder location.LocationFinder) (http.Handler,
 		rest.GetHandler(r, "/story/{storyID}", api.ReturnStoryByID)
 		rest.DeleteHandler(r, "/story/{storyID}", api.DeleteStoryByID)
 		rest.PutHandler(r, "/story/{storyID}", api.EditStoryByID)
+		rest.PutHandler(r, "/publish", api.PublishStories)
```
Shehryar21: Is the endpoint name "/publish" fine or should it be smth else?

restapi/stories.go
```diff
@@ -377,6 +377,20 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
 	return rest.MsgStatusOK("Stories added successfully")
 }
 
+func (api api) PublishStories(w http.ResponseWriter, r *http.Request) render.Renderer {
+	var stories []models.Story
+	if err := json.NewDecoder(r.Body).Decode(&stories); err != nil {
+		return rest.ErrInvalidRequest(api.logger, "Invalid payload", err)
+	}
+	for _, story := range stories {
```
nimwijetunga, line 385: So just to be clear esentially all this is doing is updating each story object right? One thing to consider would be adding a check that only the `is_visible` field has changed but that might be a bit of work

restapi/api.go
nimwijetunga: maybe `/stories/publish` would be better

restapi/stories.go
```diff
@@ -377,6 +377,20 @@ func (api api) CreateStories(w http.ResponseWriter, r *http.Request) render.Rend
 	return rest.MsgStatusOK("Stories added successfully")
 }
 
+func (api api) PublishStories(w http.ResponseWriter, r *http.Request) render.Renderer {
```
nimwijetunga, line 380: Can we add a test for this, particularly test these two cases:

1. Changing a story that was not visible to visible
2. Changing a story that was visible to not visible

nimwijetunga (changes requested):
A few changes but looks good overall

nimwijetunga (approved):
LGTM

AbhiPrasad (approved):

PR #275, opened 2021-03-27
use postal code stuff in sseding

aspensantos:


internal/database/migrations/migrations.go
```diff
@@ -92,13 +92,13 @@ func Seed(db *gorm.DB, locationFinder location.LocationFinder) error {
 	}
 
 	// set lat long for coordinates
-	for _, story := range stories {
-		coordinates, err := locationFinder.GetCityCenter(story.CurrentCity)
+	for index, story := range stories {
```
nimwijetunga, line 95: Can you run `make seed` and make sure this works and is seeded properly

internal/location/coordinates.go
```diff
@@ -85,7 +85,7 @@ func (finder coordinatesFinder) randomizePinLocation(city string) (latitude floa
 
 func (finder coordinatesFinder) GetPostalLatitudeAndLongitude(city string, limit int64) (float64, float64, error) {
 	// query zipcode with numStoriesInCity + 1 results and take the last one
-	resp, err := http.Get(fmt.Sprintf(zipCodeURL, finder.zipCodeToken, city, limit+1))
+	resp, err := http.Get(fmt.Sprintf(zipCodeURL, finder.zipCodeToken, city, limit+2))
```
nimwijetunga, line 88: Did you test that this now works?

nimwijetunga (changes requested):
Mostly just testing

internal/database/migrations/migrations.go
aspensantos: yeah i ran it and it worked

internal/location/coordinates.go
aspensantos: yeah, tested

nimwijetunga (approved):
LGTM

PR #277, opened 2021-03-31
fix: upload story button working

DinuWije:


hanlinc27 (approved):
Upload works fine for me 👍🏻  (but it also did before when making the button `type="submit" `and defining `onSubmit = {handleSubmit}` ) so not entirely sure why it wouldn't be working?

PR #280, opened 2021-04-03
feat: Add frontend auth

AbhiPrasad:
Dependent on https://github.com/uwblueprint/shoe-project/pull/266.

This PR adds frontend auth with https://www.npmjs.com/package/react-google-login.

I didn't move any routes to be protected because of ease of development reasons, but I tested and it works out 👍


ui/src/hooks/auth.tsx
```diff
@@ -1,65 +1,145 @@
...
-import { Redirect, Route, RouteProps } from "react-router-dom";
+import {
+  GoogleLoginResponse,
+  GoogleLoginResponseOffline,
+  useGoogleLogin,
+  useGoogleLogout,
+} from "react-google-login";
 
-import { User } from "../types";
+const CLIENT_ID =
```
aspensantos: is this okay here or do we want to put this in the .env?

ui/src/pages/admin/AllStories.tsx
```diff
@@ -39,7 +39,7 @@ const StyledAppBar = styled(AppBar)`
 `;
 
 export const AllStories: React.FC = () => {
-  const { data: allStories, error } = useSWR<Story[]>("/api/stories");
+  const { data: allStories, error } = useSWR<Story[]>("/api/stories", {});
```
aspensantos: what does this do?

AbhiPrasad: ya fixed.

ui/src/hooks/auth.tsx
AbhiPrasad: Great point, I'll pull this in from .env

nimwijetunga (approved):
This looks good to me, but with my limited react experience you should get someone else to approve as well

ui/src/App.tsx
```diff
@@ -27,20 +27,11 @@ function App(): JSX.Element {
               <ShoeMap />
             </Route>
             <Route exact path="/login">
-              <Login login />
+              <Login />
```
hanlinc27, line 30: Login gives me this console error:
Screen Shot 2021-04-08 at 5 55 07 PM
And Admin shows this error:
Screen Shot 2021-04-08 at 5 53 46 PM

@aspensantos @nimwijetunga did this not show up for you? If not, maybe it's just my side?

hanlinc27 (commented):
`/login` and `/admin` are both hanging for me and don't render anything?

The request seems to have gone through fine though
image

nimwijetunga:
Good point I didn't try running it locally before but yeah im seeing those same issues as well

ui/src/App.tsx
nimwijetunga: Good point I didn't try running it locally before but yeah im seeing those same errors as well

AbhiPrasad: Fixed, for some reason the imports messed up.

AbhiPrasad:
Alright, fixed things up. We are hardcoding the google client ID because environmental variables are a pain with Parcel.

nimwijetunga (changes requested):
I tried this out locally it works fine except for one case:

- Log in as an invalid user (get an error and says incorrect email which is good)
- When i try click on sign in again it never gives me the chance to reenter my email (just constantly keeps pinging the backend with the previous email and giving me invalid user)

I think google is caching something (since it was a valid login, but rejected by our backend) because when i clear all the cookies in the browser it works fine (chrome btw)

nimwijetunga:
I dug a bit more into this issue it seems that we need to log the user out of google when we ask them to sign in again (i.e calling this url works https://www.google.com/accounts/Logout?continue=https://appengine.google.com/_ah/logout?continue=http://localhost:8900/login)

AbhiPrasad:
Alright I fixed that, give it a try now.

nimwijetunga:
hmm still seems to have the same issue don't know if its an issue locally or not (are you seeing this problem as well)?

nimwijetunga (approved):
Tested LGTM

AbhiPrasad:
This is now ready to review again

ui/src/__tests__/pages/Admin/AllStories.test.ts
```diff
@@ -4,7 +4,7 @@ import {
   Action,
   allStoriesReducer,
   INIT_STATE,
-} from "../../pages/AllStories/reducer";
+} from "../../../pages/Admin/AllStories/reducer";
```
hanlinc27, line 7:
```suggestion
} from "../../../pages/admin/AllStories/reducer";
```
is that what's breaking the test import?

ui/src/App.tsx
hanlinc27, line 30: Is there a variable I'm missing? Getting this in my make error:
image

Also this is expected behaviour for `/login` and `admin` - my blueprint email should be an invalid sign in?
bcd47521-aea2-41de-a27b-34cbc6e1bdfe

hanlinc27 (commented):
Backend API Error:
image

ui/src/App.tsx
hanlinc27: Actually wait that shouldn't be expected behaviour because it looks like `auth.go` should validate uwblueprint.org emails.

ui/src/__tests__/pages/Admin/AllStories.test.ts
AbhiPrasad: Yup fixed

ui/src/App.tsx
AbhiPrasad: Huh weird @nimwijetunga might have more context here.

nimwijetunga: @hanlinc27 copy the newest env file pinned to the shoe project slack channel. The error your getting is because you have a wrong client id (old env file has wrong credentials)

hanlinc27: ooh yea you're right, it works now we good

ui/src/pages/Admin/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  if (!googleLoaded) {
+    return (
+      <Center>
+        <CircularProgress />
+      </Center>
+    );
+  }
+
+  if (auth === undefined) {
+    return <Redirect to="/login" />;
```
hanlinc27, line 32: I don't get redirected to login, it just infinitely shows the loading icon at the moment when I'm not logged in.

hanlinc27 (approved):
`/login` works for me now w/ my blueprint email! Just a quick question about the `/admin` route

ui/src/pages/Admin/index.tsx
AbhiPrasad: I'll investigate this, for now merging it in.

PR #282, opened 2021-04-03
feat: Search Stories Frontend

Shehryar21:
### Description
This implements search for all three tabs on allstories. It should exclude searching for any parameters not displayed on table cells (i.e. content, image_url etc.) Toggling should persist before search and can also be toggled afterwards. 

image


ui/src/pages/AllStories/index.tsx
```diff
@@ -208,6 +227,56 @@ export const AllStories: React.FC = () => {
     dispatch({ type: "SET_TAB_VALUE", newValue });
   };
 
+  const requestSearchHelper = (row, searchedVal: string) => {
+    let doesExist = false;
+    Object.keys(row).forEach((prop) => {
+      //Exclude search for StoryView members not displayed on table cells
+      const excludedParameters =
```
hanlinc27: @nimwijetunga We don't want it to show stories that may include the search keywords inside of its content right? I.e. they search "Toronto" and a story mentions Toronto in their details so then it'll show up in the valid stories

nimwijetunga: Yeah lets exclude any fields that are not directly visible on the table

ui/src/pages/AllStories/reducer.ts
```diff
@@ -19,14 +23,21 @@ export type Action =
   | { type: "HANDLE_CHECKED"; e: React.ChangeEvent; story: StoryView }
   | { type: "SET_ORDERING"; order: "asc" | "desc"; orderBy: string }
   | { type: "SET_TABLE_DATA"; data: StoryView[] }
-  | { type: "SET_TAB_VALUE"; newValue: number };
+  | { type: "SET_TAB_VALUE"; newValue: number }
+  | { type: "SET_CHANGED_VISIBILITY"; data: StoryView[] }
```
DinuWije, line 27: consider changing the input variable name for readability (ie. instead of "data", "newVisibility")

ui/src/pages/AllStories/index.tsx
```diff
@@ -552,6 +623,13 @@ export const AllStories: React.FC = () => {
           <StyledEmptyMessage> No pending changes! </StyledEmptyMessage>
         )}
       </AllStoriesTabs>
+      <StoryDrawer
+        story={clickedStory}
+        onClose={() => setClickedStory(undefined)}
+        onClickEditStory={() => {
+          console.log("TODO: Route to edit page");
```
DinuWije, line 656: Did you try adding this route with a Redirect?

DinuWije (commented):
Looks good to me, but pls get a second look from one of the other devs

ui/src/pages/AllStories/index.tsx
hanlinc27: That one's to route to edit stories which needs to get merged in from your PR?

ui/src/pages/AllStories/reducer.ts
hanlinc27: hmm I think we should just keep it all as data for consistency? The type should be explanatory for what the action does

ui/src/pages/AllStories/index.tsx
```diff
@@ -229,9 +248,60 @@ export const AllStories: React.FC = () => {
...
     dispatch({ type: "SET_TAB_VALUE", newValue });
   };
 
+  const requestSearchHelper = (row, searchedVal: string) => {
+    let doesExist = false;
+    Object.keys(row).forEach((prop) => {
+      //Exclude search for StoryView members not displayed on table cells
+      const excludedParameters =
+        prop !== "image_url" &&
+        prop !== "video_url" &&
```
nimwijetunga, line 261: Lets also hide the id column (i.e when i search `13` story with id 13 will also show up). Maybe we can remove the id column from the table itself (i think anna wanted this in the first place).

nimwijetunga (approved):
Tested it out locally and seems to work fine for me

PR #283, opened 2021-04-03
changed allstories to return visible and non-visible stories

Shehryar21:
just hitting api/stories would return all visible and non-visible stories. To get only visible or non-visible, pass in visibility query param

### DIFF
--- a/restapi/endpoints_test.go
+++ b/restapi/endpoints_test.go
@@ -317,7 +317,7 @@ func (suite *endpointTestSuite) TestGetAllStories() {
 
 	//Verify response matches
 	response.Schema(mock)
-	response.Object().Value("payload").Array().Length().Equal(4)
+	response.Object().Value("payload").Array().Length().Equal(7)
 
 	response.Object().Value("payload").Array().Element(0).Object().Value("author_first_name").Equal("John")
 	response.Object().Value("payload").Array().Element(0).Object().Value("author_last_name").Equal("Doe")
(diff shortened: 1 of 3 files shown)


nimwijetunga (approved):
LGTM

restapi/endpoints_test.go
```diff
@@ -482,6 +482,7 @@ func (suite *endpointTestSuite) TestGetVisibleStoriesWithTags() {
 	suite.db.Create(&jsonStories)
 	suite.db.Create(&jsonTags)
 	var response = suite.endpoint.GET("/stories").
+		WithQuery("visibility", true).
```
AbhiPrasad, line 485: Can we add a test response test for `WithQuery("visibility", false).`?

AbhiPrasad: Also shouldn't there be three modes in this enum? We can get visible stories, not visible stories or all stories. We should then have `visibility=true/false` for the first two conditions, and then if they don't pass in anything at all we can return all stories regardless of visibility

AbhiPrasad (changes requested):

restapi/endpoints_test.go
Shehryar21: yes we already have a test for invisible stories. Look at the func TestGetAllInvisibleStories().

Shehryar21: and thats exactly how it is. If there is no query param visibility, then we return all stories, otherwise it returns either visible or non-visible depending on whether visibility was passed in as true or false

AbhiPrasad: Ah ok I'm a dummy, that's my bad for just skimming and not actually reading lol.

AbhiPrasad (approved):

PR #281, opened 2021-04-03
feat: add max file size limit on upload

hanlinc27:
### Description

Set limit on file size upload for #279 . To test, go to `/upload` download a file > 7MB (i.e. https://effigis.com/en/solutions/satellite-images/satellite-image-samples/) Should get the following error: 
image
image

Closes #278

PR #285, opened 2021-04-07
Upload & Edit Stories

DinuWije:
## Major Changes 
-  Refactored Hanlin's code for uploading a story into it's own component 
- Prefill story details for stories being edited (endpoint is "/edit/storyID")
- Added Autocomplete selectors for country and city fields
- Added option to add a new country to the list (and update the backend) 
- Merged in Aspen's PR for upload navbar

# TODO
- Redirect to Hanlin's upload/edit complete page 
- lots of QA lol


cmd/server.go
```diff
@@ -39,8 +39,10 @@ var (
 			server.Router.Get("/*", func(w http.ResponseWriter, r *http.Request) {
 				if _, err := os.Stat(dir + r.RequestURI); os.IsNotExist(err) {
 					http.StripPrefix(r.RequestURI, fs).ServeHTTP(w, r)
+					logger.Info("here 1")
```
AbhiPrasad: 🔥 
```suggestion
```

cmd/server.go
```diff
@@ -39,8 +39,10 @@ var (
...
 				} else {
 					fs.ServeHTTP(w, r)
+					logger.Info("here 2")
```
AbhiPrasad:
```suggestion
```

ui/src/components/UploadStory.tsx
```diff
@@ -0,0 +1,855 @@
+import {
```
AbhiPrasad: ok plz split this component, and keep it with the `Edit` page instead of storing it under `src/components/`

```
pages
  ...
  Edit -> folder
    index.tsx -> has the page component
    UploadStory.tsx -> upload story component
    ListboxComponent.tsx -> The react window logic
    types.ts -> has the types
    ... any other files you find needed
  ...
 ```

File: ui/src/components/UploadStory.tsx [hunk 4]
```diff
@@ -0,0 +1,855 @@
...
+    currentStory.current_city.toUpperCase()
+  );
+  const [autocompleteAuthorCountry, setAutocompleteAuthorCountry] = useState(
+    currentStory.author_country
+  );
+  const [addedCountry, setAddedCountry] = useState("");
+  const [newImage, setNewImage] = useState(currentStory.image_url);
+  const [disabled, setDisabled] = useState(false);
+  //handleSubmit component states
+  const [dialogOpenState, setDialogOpenState] = useState(false);
```
AbhiPrasad: multiple use states is a code smell. Let's refactor this into `useReducer` (see https://github.com/uwblueprint/shoe-project/pull/267/files#diff-5c3739858bec843543a6cf139cc3b2e3febe6fe751e55dbb2c20daf08f8d76b2R1 for an example)

ui/src/components/UploadStory.tsx
```diff
@@ -0,0 +1,855 @@
...
+  const { data: countries, error: errorCountries } = useSWR<string[]>(
+    "/api/countries"
+  );
+
+  const yearArray = [];
+  let year: number = new Date().getFullYear();
+  for (let i = 30; i > 0; i--) {
+    yearArray.push(year);
+    year -= 1;
+  }
```
AbhiPrasad: A little JS magic :)

```suggestion
const startYear = new Date().getFullYear();
const yearArray = Array.from({ length: 30 }, (_, i) => startYear - i)
```

AbhiPrasad (changes requested):
1st pass review

ui/package.json
```diff
@@ -23,22 +23,24 @@
...
-    "react": ">=15",
-    "react-dom": ">= 16.8",
+    "react": "^15.3.0 || ^16.0.0-alpha",
+    "react-dom": "^15.3.0 || ^16.0.0-alpha",
     "react-is": "^16.13.1",
     "react-leaflet": "^2.0.0",
     "react-leaflet-control": "^2.1.2",
     "react-leaflet-markercluster": "^2.0.0",
     "react-router-dom": "^5.2.0",
+    "react-virtualized": "^9.22.3",
```
hanlinc27: I don't think you're using `react-virtualized` anywhere only `react-window` so this can be removed

ui/package.json
```diff
@@ -23,22 +23,24 @@
...
     "@material-ui/styles": "^4.11.3",
     "leaflet": "^1.0.2",
     "leaflet.markercluster": "^1.4.1",
     "material-ui-dropzone": "^3.5.0",
-    "mui-virtualized-table": "3.0.0-6",
+    "mui-virtualized-table": "^2.2.3",
     "prop-types": "^15.7.2",
-    "react": ">=15",
-    "react-dom": ">= 16.8",
+    "react": "^15.3.0 || ^16.0.0-alpha",
```
hanlinc27: Can we please keep it as  
`"react": "^16.8.0", "react-dom": "^16.8.0",`

ui/package.json
```diff
@@ -23,22 +23,24 @@
...
-    "@material-ui/core": "4.11.1",
+    "@material-ui/core": "^4.11.3",
     "@material-ui/icons": "^4.0.0",
     "@material-ui/lab": "^4.0.0-alpha.56",
     "@material-ui/styles": "^4.11.3",
     "leaflet": "^1.0.2",
     "leaflet.markercluster": "^1.4.1",
     "material-ui-dropzone": "^3.5.0",
-    "mui-virtualized-table": "3.0.0-6",
+    "mui-virtualized-table": "^2.2.3",
```
hanlinc27: Please bump it up to` "mui-virtualized-table": "3.0.0-6" `before merge, the older version has some bugs that are incompatible with our other MUI components

ui/src/components/UploadStory.tsx
```diff
@@ -0,0 +1,855 @@
...
+}
+
+const OuterElementContext = React.createContext({});
+
+const OuterElementType = React.forwardRef<HTMLDivElement>((props, ref) => {
+  const outerProps = React.useContext(OuterElementContext);
+  return <div ref={ref} {...props} {...outerProps} />;
+});
+
+function useResetCache(data: any) {
```
hanlinc27:
```suggestion
function useResetCache(data: number) {
```

ui/src/components/UploadStory.tsx
```diff
@@ -0,0 +1,855 @@
...
+  React.useEffect(() => {
+    if (ref.current != null) {
+      ref.current.resetAfterIndex(0, true);
+    }
+  }, [data]);
+  return ref;
+}
+
+// Adapter for react-window
+const ListboxComponent = React.forwardRef<HTMLDivElement>(
```
hanlinc27: Can we separate this `ListboxComponent` into another file?

ui/src/components/UploadStory.tsx
```diff
@@ -0,0 +1,855 @@
...
+  const handleChange = (
+    event?: React.ChangeEvent<{ name?: string; value: unknown }>
+  ) => {
+    const name = event.target.name as keyof typeof formInput;
+    setFormInput({
+      [name]: event.target.value,
+    });
+  };
+
+  const handleDialogClose = () => {
```
hanlinc27: Since we're going to be implementing the new page on successful upload, let's remove all the success dialog code:
`handleDialogClose()`
`handleDialogAgree()`
`hadleDialogDisagree()`

 etc

ui/src/components/UploadStory.tsx
```diff
@@ -0,0 +1,855 @@
...
+      ["author_country"]: autocompleteAuthorCountry,
+    });
+    setAddedCountry(autocompleteAuthorCountry);
+    setAuthorCountry(autocompleteAuthorCountry);
+    countries.push(autocompleteAuthorCountry);
+  };
+
+  const addNewCountry = ({ children, ...other }) => (
+    <AddCountryPaper {...other}>
+      {countries.filter(
```
hanlinc27: Can we use strict equality/inequality checks in these comparators please

ui/src/components/UploadStory.tsx
```diff
@@ -0,0 +1,855 @@
...
+                    React.HTMLAttributes<HTMLElement>
+                  >
+                }
+                renderGroup={renderGroup}
+                forcePopupIcon
+                freeSolo
+                name="current_city"
+                id="select-label-city"
+                value={currentCity != "" ? currentCity : null}
+                //TODO: double check that this onChange method works
```
hanlinc27: Seems to work for me. Can delete comment

ui/src/components/UploadStory.tsx
```diff
@@ -0,0 +1,855 @@
...
+              <UploadStoriesHeading>Multimedia</UploadStoriesHeading>
+              <ImageContainer>
+                {newImage == "" ? (
+                  <div>
+                    <UploadLabelsText>Add Image</UploadLabelsText>
+                    <StyledDropzoneArea
+                      showFileNames
+                      acceptedFiles={["image/*"]}
+                      filesLimit={1}
+                      dropzoneText={"Drag image here or select from device"}
```
hanlinc27: Make sure to merge in updates from https://github.com/uwblueprint/shoe-project/pull/281 for image size

ui/src/components/UploadStory.tsx
```diff
@@ -0,0 +1,855 @@
...
+  const [newImage, setNewImage] = useState(currentStory.image_url);
+  const [disabled, setDisabled] = useState(false);
+  //handleSubmit component states
+  const [dialogOpenState, setDialogOpenState] = useState(false);
+  const [uploadErrorState, setErrorOpen] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const [isDrawerOpen, setIsDrawerOpen] = useState(false);
+
+  const { data: countries, error: errorCountries } = useSWR<string[]>(
+    "/api/countries"
```
hanlinc27: Should this endpoint be returning an empty array right now?

ui/src/components/UploadStory.tsx
```diff
@@ -0,0 +1,855 @@
...
+          >
+            <StyledButton
+              disabled={!hasAllRequiredFields}
+              onClick={() => setIsDrawerOpen(true)}
+              color="primary"
+              variant="outlined"
+            >
+              Preview
+            </StyledButton>
+            <StyledButton
```
hanlinc27: The button is still disabled when I have all my fields filled out during upload. 
image

hanlinc27 (changes requested):
✏️ 📝

cmd/server.go
DinuWije: 😅 thanks

ui/src/components/UploadStory.tsx
DinuWije: No, are u getting an empty array when you try it?

ui/src/components/UploadStory.tsx
DinuWije: Will do. What exactly is going on here (is there a name for it)? I can tell what Array.from does but not how the interior works.

ui/src/components/UploadStory.tsx
```diff
@@ -0,0 +1,855 @@
...
+        DeletedAt: {
+          Time: "",
+          Valid: true,
+        },
+      };
+      return storyFromData;
+    }
+    return undefined;
+  }, [isDrawerOpen, formInput]);
+
```
DinuWije: @hanlinc27 @AbhiPrasad @aspensantos is it okay to leave this warning alone ^^? The code is from Aspen's commit so I'm not too sure what the problem is

ui/src/components/UploadStory.tsx
DinuWije: Just wondering, what is the benefit of that in this case considering we know that autocompleteAuthorCountry is a string? Is it just best practice?

ui/src/components/UploadStory.tsx
DinuWije: Hm, did you upload an image as well? Or if you added a country not on the list did you make sure to click the add button?

ui/src/components/UploadStory.tsx
DinuWije: Will do, but tbh I need to ask you and @AbhiPrasad about this because I'm not sure if it's actually doing anything lol

ui/src/components/UploadStory.tsx
DinuWije: nice catch, ty

ui/src/components/UploadStory.tsx
DinuWije: Just to clarify, index.tsx is the page for edit story ONLY right (basically just move the current page in there)?

ui/src/components/UploadStory.tsx
hanlinc27: Yeah for readability if someone came into our codebase and didn't know they were the same type and also for consistency sake since we strict equality check everywhere else.

ui/src/components/UploadStory.tsx
AbhiPrasad: yup

ui/src/components/UploadStory.tsx
AbhiPrasad: `Array.from` creates an array from an "array-like" object. In this case the object defines a length, so it can be considered array like. The second argument calls a callback for each element in the array. Here we just to the same logic you did in the for loop.

ui/package.json
DinuWije: Lol kinda confused, I accidentally deleted react-window instead of virtualized but everything seems to work normally... will ask u about this on Saturday.

ui/src/data/cities.tsx
```diff
@@ -0,0 +1,6645 @@
+export const citiesList = [
```
DinuWije, line 1: Forgot to use git mv, dont kill me Abhijeet 😩

ui/src/pages/Edit/UploadStory.tsx
```diff
@@ -0,0 +1,721 @@
+import {
```
DinuWije: @AbhiPrasad i used git mv but it still has massive diffs, im guessing I did it wrong? lmk what you think

ui/src/components/UploadStory.tsx
hanlinc27: Ah had to run `make seed` it's working fine now

ui/src/components/UploadStory.tsx
hanlinc27: Working fine now, I think it was b/c countries wasn't there

ui/src/pages/Edit/Listbox.tsx
```diff
@@ -0,0 +1,91 @@
...
+  React.useEffect(() => {
+    if (ref.current != null) {
+      ref.current.resetAfterIndex(0, true);
+    }
+  }, [data]);
+  return ref;
+}
+
+// Adapter for react-window
+export const ListboxComponent = React.forwardRef<HTMLDivElement>(
```
hanlinc27: Commenting out this component inside of `UploadStory` doesn't seem like it actually changes/breaks anything?

ui/src/components/UploadStory.tsx
DinuWije: Ah cool, thanks!

ui/src/pages/Edit/Listbox.tsx
```diff
@@ -0,0 +1,91 @@
...
+          >
+            {renderRow}
+          </VariableSizeList>
+        </OuterElementContext.Provider>
+      </div>
+    );
+  }
+);
+
+export const renderGroup = (params: AutocompleteRenderGroupParams) => [
```
AbhiPrasad:
```suggestion
export const renderGroup = (params: AutocompleteRenderGroupParams): JSX.Element[] => [
```

hanlinc27:
```suggestion
export const renderGroup = (params: AutocompleteRenderGroupParams)): React.ReactNode[] => [
```

hanlinc27 (approved):
`Edit`, `Upload` all work for me and I can see the changes in the allstories table :)

Edit: @DinuWije  Actually can you route the success page to the correct one before merging this?

ui/src/pages/Edit/Listbox.tsx
DinuWije: It doesn't break anything, but just from messing around with the Cities autofill component I find that including the listbox SEEMS to make it go faster/have less lag. I'm not sure if this is just a placebo on my end, and if you think the listbox is not needed we can get rid of it.

hanlinc27: LOL yeah I mean it is really smooth, let's just go with it?

DinuWije: sounds good, ty

ui/src/pages/Edit/Listbox.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { useTheme } from "@material-ui/core/styles";
+import useMediaQuery from "@material-ui/core/useMediaQuery";
+import { AutocompleteRenderGroupParams } from "@material-ui/lab/Autocomplete";
+import * as React from "react";
+import { ListChildComponentProps, VariableSizeList } from "react-window";
+
+const LISTBOX_PADDING = 8; // px
+
+function renderRow(props: ListChildComponentProps) {
+  const { data, index, style } = props;
```
AbhiPrasad:
```suggestion
function renderRow({ data, index, style }: ListChildComponentProps) {
```

ui/src/pages/Edit/UploadStory.tsx
AbhiPrasad: ya its fine whatever. Just rack up those contributions :)))

ui/src/pages/Edit/UploadStory.tsx
```diff
@@ -0,0 +1,721 @@
...
+        UpdatedAt: "",
+        DeletedAt: {
+          Time: "",
+          Valid: true,
+        },
+      };
+      return storyFromData;
+    }
+    return undefined;
+  }, [state.isDrawerOpen, formInput]);
```
AbhiPrasad: We should add the dependencies, otherwise we risk not invalidating the cache properly.

ui/src/pages/Edit/UploadStory.tsx
```diff
@@ -0,0 +1,721 @@
...
+  ) => {
+    const name = event.target.name as keyof typeof formInput;
+    setFormInput({
+      [name]: event.target.value,
+    });
+  };
+
+  const storySubmitDialog = (result) => {
+    dispatch({ type: "SET_LOADING_STATUS", loadingStatus: false });
+    dispatch({ type: "SET_BUTTON_DISABLE", status: false });
```
AbhiPrasad: Refactor into a single dispatch plz.

ui/src/pages/Edit/UploadStory.tsx
```diff
@@ -0,0 +1,721 @@
...
+
+  const addCountryButtonPressed = (autocompleteAuthorCountry: string) => {
+    setFormInput({
+      ["author_country"]: autocompleteAuthorCountry,
+    });
+    dispatch({ type: "ADD_COUNTRY", newCountry: autocompleteAuthorCountry });
+    dispatch({
+      type: "SET_AUTHOR_COUNTRY",
+      country: autocompleteAuthorCountry,
+    });
```
AbhiPrasad: Single dispatch plz

ui/src/pages/Edit/UploadStory.tsx
```diff
@@ -0,0 +1,721 @@
...
+  currentStory,
+  bio,
+}: StoryProps) => {
+  const { data: tagOptions, error } = useSWR<string[]>("/api/tags");
+  const { data: countries, error: errorCountries } = useSWR<string[]>(
+    "/api/countries"
+  );
+  const [state, dispatch] = useReducer(
+    uploadStoryReducer,
+    INIT_STATE(currentStory)
```
AbhiPrasad: If this is a function, lets not make it all uppercase. 
```suggestion
    get_init_state(currentStory)
```

ui/src/pages/Edit/UploadStory.tsx
```diff
@@ -0,0 +1,721 @@
...
+  const addCountryButtonPressed = (autocompleteAuthorCountry: string) => {
+    setFormInput({
+      ["author_country"]: autocompleteAuthorCountry,
+    });
+    dispatch({ type: "ADD_COUNTRY", newCountry: autocompleteAuthorCountry });
+    dispatch({
+      type: "SET_AUTHOR_COUNTRY",
+      country: autocompleteAuthorCountry,
+    });
+    countries.push(autocompleteAuthorCountry);
```
AbhiPrasad: mutable state is icky. What we should do is have two layers.

```
// handles data fetching
function UploadStoryWrapper() {
  const { data: tagOptions, error } = useSWR<string[]>("/api/tags");
  const { data: countries, error: errorCountries } = useSWR<string[]>(
    "/api/countries"
  );

 // handle error states
 // ...

  if (tagOptions === undefined || countries === undefined) {
    return some loading indicator
  }
  
  UploadStories component guaranteed to have filled tagsOptions and countries
  return <UploadStory tagOptions={tagOptions} countries={countries};
}
```

We can then pass in `countries` into the reducer state -> `INIT_STATE(countries, currentStory)`. Instead of calling `countries.push()`, we dispatch an action to push countries:

```
{
  ...
  countries: [...state.countries, action.newCountry]
}
```

ui/src/pages/Edit/UploadStory.tsx
```diff
@@ -0,0 +1,721 @@
...
+  };
+
+  const addNewCountry = ({ children, ...other }) => (
+    <AddCountryPaper {...other}>
+      {countries.filter(
```
AbhiPrasad: Extract to variable plz.

ui/src/pages/Edit/UploadStory.tsx
```diff
@@ -0,0 +1,721 @@
...
+                <StyledAutocomplete
+                  color="Primary"
+                  loading={!countries}
+                  options={countries || []}
+                  forcePopupIcon
+                  name="author_country"
+                  id="select-label-country"
+                  freeSolo
+                  // if user adds a new country display it as the default value, if not, look for prefilled value (edit stories)
+                  // prettier-ignore
```
AbhiPrasad: Why prettier-ignore?

ui/src/pages/Edit/UploadStory.tsx
```diff
@@ -0,0 +1,721 @@
...
+                forcePopupIcon
+                freeSolo
+                name="current_city"
+                id="select-label-city"
+                value={state.currentCity != "" ? state.currentCity : null}
+                onChange={(_, newValue) => {
+                  setFormInput({
+                    ["current_city"]: newValue,
+                  });
+                  dispatch({ type: "SET_CURRENT_CITY", city: newValue });
```
AbhiPrasad: Maybe we can combine these into one big reducer state?

AbhiPrasad (changes requested):
Part 2/3, we almost there!

ui/src/pages/Edit/UploadStory.tsx
DinuWije: yessir, i'll try again if it comes up.

ui/src/pages/Edit/UploadStory.tsx
DinuWije: it looks ugly with prettier running on it lol, can delete the comment tho

ui/src/pages/Edit/UploadStory.tsx
DinuWije: I'm not sure what you mean by this. Do u mean refactoring the formInput reducer into the main reducer component?

AbhiPrasad: yup, so there is only one `useReducer()` call.

ui/src/pages/Edit/UploadStory.tsx
DinuWije: Okay so I've come across a problem with this suggested solution: We can't return any TSX before all the hooks are declared. If we return at the `if(tagOptions === undefined ...)`, the hooks below it will be conditionally instantiated. We can't move those hooks up because parts of them rely on the state reducer (which in your suggested edit would be AFTER the if statement) --> thus we have a cycle where we are forced to conditionally create hooks.  This is also the reason the error check for the useSWR methods appears right before the final return statement. 

One solution I thought of for this: set all of our TSX to a variable and change that variable as needed, then return it at the very end of the function. Something like:
```
let returnTSX = <div/>
if(tagOptions === ....) returnTSX = <div>IF</div>;
else returnTSX=<div>ELSE</div>;
``` 
(along those lines) 
However, the idea of setting the return TSX to a variable seems very messy to me and doesn't work quite as I thought it would when I was messing around with it. 
Let me know what you think.

AbhiPrasad: No the hooks wouldn't be conditionally rendered because they would be in two different components. Essentially we have a data fetching component, and we have a view layer component.

```ts
function UploadStoryWrapper() {
  // fetch data with SWR hooks
  // if data error return error
  // if data undefined return loading
  // if data defined return <UploadStory tagOptions={tagOptions} countries={countries} />
}

...

function UploadStory({ tagOptions, countries }) {
  const [state, dispatch] = useReducer(reducer, get_init_state(countries, currentStory));
  // do whatever hook calls you want here

  // return final product to get rendered
}
```

AbhiPrasad:
I merged in https://github.com/uwblueprint/shoe-project/pull/280, which might be a hefty rebase.

Take a look at the new page layout https://github.com/uwblueprint/shoe-project/tree/main/ui/src/pages.

`src/pages/Edit` -> `src/pages/Admin/Edit`

ui/src/pages/Edit/UploadStory.tsx
DinuWije: got it, thanks for clarifying!

ui/src/pages/Edit/UploadStory.tsx
DinuWije: Okay so I tried going at this for a bit, but I'm having a tough time making it work. Here's what I'm doing right now: 
- created a formInput type in types.ts, and then added `formInput` as one of the fields in the State interface of my reducer. 
- working on switching the `setFormInput[]` methods in UploadStory to dispatches
Where I'm stuck: 
let's say I want to change just the `Title` field of my formInput. Inside the case statements of my reducer, I have the case: 
```
case "title": {
      return{
        ...state,
        formInput.title: action.formData,
      }
    }
```

But this gives me errors as it doesn't let me edit formInput this way. 

pls lmk if this is the right approach, or if there's a better way to do this. Sorry for the long comment!

ui/src/pages/Edit/index.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from "react";
+import { useParams } from "react-router-dom";
+import useSWR from "swr";
+
+import { Story } from "../../types/index";
+import { UploadStoryWrapper } from "./UploadStoryWrapper";
+
+export const Edit: React.FC = () => {
+  const { id } = useParams();
+  const { data: story, error } = useSWR<Story>(`/api/story/${id}`);
```
AbhiPrasad: Can we just make the `Edit` component have all the datafetching? We don't need a `UploadStoryWrapper` then

```ts
export const Edit: React.FC = () => {
  const { id } = useParams();
  const { data: story, error: errorStory } = useSWR<Story>(`/api/story/${id}`);
  const { data: tagOptions, error: errorTags } = useSWR<string[]>("/api/tags");
  const { data: countries, error: errorCountries } = useSWR<string[]>(
    "/api/countries"
  );

  if (errorStory) {
    return <InfoDiv>Error fetching story!</InfoDiv>;
  }
  if (errorTags) {
    return <InfoDiv>Error fetching tags!</InfoDiv>;
  }
  if (errorCountries) {
    return <InfoDiv>Error fetching countries array!</InfoDiv>;
  }

  if (tagOptions === undefined || countries === undefined || story === undefined) {
    return <StyledProgress />;
  }

  return (
    <UploadStory
      id={id}
      currentStory={story}
      bio={story.author.bio}
      tagOptions={tagOptions}
      countries={countries}
    />
  );

}
```

AbhiPrasad (changes requested):

ui/src/pages/Edit/index.tsx
DinuWije: I feel like that would be more redundant because we would have to copy the same code over to the Upload component as well. Having the wrapper allows us to fetch the data common to both pages via one component. If you think the way you suggested is still more efficient, pls let me know and I can make the change.

nimwijetunga (approved):
major functionality of edit stories work there are a few outstanding issues (in terms of additions) but its prob better doing in a separate pr since this one is already pretty big

nimwijetunga:
I will merge this pr in so that we can begin E2E testing on tuesday. If there are any further comments leave them here and @DinuWije can address them if there is enough time

PR #286, opened 2021-04-07
Added login link to homepage

DinuWije:
## Changes
- Added link to login page from the homepage 
image

### DIFF
--- a/ui/src/components/HelpAccordion.tsx
+++ b/ui/src/components/HelpAccordion.tsx
@@ -4,6 +4,7 @@ import AccordionSummary from "@material-ui/core/AccordionSummary";
 import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
 import * as React from "react";
 import { useState } from "react";
+import { Link } from "react-router-dom";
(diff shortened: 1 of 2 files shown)


ui/src/components/HelpAccordion.tsx
```diff
@@ -83,10 +84,16 @@ const StyledBox = styled.div`
...
   title: string;
   body: string;
+  login: boolean;
 }
 
-export function HelpAccordion({ title, body }: HelpDrawerProps): JSX.Element {
+export function HelpAccordion({
+  title,
+  body,
+  login,
```
AbhiPrasad: let's make this false by default so we don't have to keep repeating `login={false}`

```suggestion
  login = false,
```

AbhiPrasad (approved):

ui/src/components/HelpAccordion.tsx
DinuWije: okay, made login into an optional field and set it to false by default

hanlinc27 (approved):

PR #288, opened 2021-04-10
fix: Pending table bug, general all stories component stylings

hanlinc27:
### Description
Some small changes in this PR: 

- Fixed the problem in pending table where it was always reading state as false (changed the visibleState array to number so we can properly check with `.includes`
- Styled switches, add-icon, and remove-icon according to [Figma ](https://www.figma.com/file/uVqnweL2ApqvPHWzQYmZ1t/Dashboard?node-id=1083%3A46971)
- Added number of current stories in each tab

### Screenshots
image
image


nimwijetunga (approved):
Tried it out locally and looks good to me!

PR #290, opened 2021-04-10
feat: upload button, empty states, leave page warning

### LINKED ISSUE #289: feat: Additional allstories table components - warning, upload button, empty states
Upload Button: 
image
Empty States:
image

hanlinc27:
Closes #289 
### Description 
Some small changes again: 

- Empty state for visible stories and all stories table
- Upload New Story Button
- Warning on page route change when user tries to upload new story 

### Upload Button and Empty States
Screen Shot 2021-04-10 at 2 48 24 PM
Screen Shot 2021-04-10 at 2 50 31 PM
### Leave Page Warning 
image


ui/src/pages/AllStories/index.tsx
```diff
@@ -334,226 +367,251 @@ export const AllStories: React.FC = () => {
...
+            <StyledEmptyMessage> No Stories Here!</StyledEmptyMessage>
+            <StyledSubEmptyMessage>
+              {" "}
+              To add a story, click “ Upload Story” on the top right.{" "}
+            </StyledSubEmptyMessage>
+          </>
+        )}
+      </AllStoriesTabs>
+      <AllStoriesTabs value={state.tabValue} index={1}>
+        {state.visibleTableState.filter((story) => story.is_visible).length !==
```
AbhiPrasad: let's move this boolean into it's own variable?

ui/src/pages/AllStories/index.tsx
```diff
@@ -306,25 +324,40 @@ export const AllStories: React.FC = () => {
   if (!allStories) return <div>Loading all stories table..</div>;
   return (
     <>
+      <Prompt
+        message={(location) =>
+          `Are you sure you want to go to ${location.pathname}? Your changes on this page will not be saved.`
```
AbhiPrasad: Wait shouldn't we only conditionally render this prompt? Right now we render it everytime, regardless of the page condition.

AbhiPrasad (approved):

ui/src/pages/AllStories/index.tsx
hanlinc27: Set a boolean variable that becomes true whenever a switch or checkbox gets toggled?

nimwijetunga (changes requested):
2 things:

1. When i toggle the visibility of some stories, click upload, then click no on the prompt, all stories are set to is_visble=true. The weird part is published stories is fine (it still says all stories which have changed) but all stories and visible stories revert back to original form (where all stories are set to visible)
2. Can we also add this prompt when we click on a story then click edit story (so basically on the edit story button)

ui/src/pages/AllStories/index.tsx
AbhiPrasad: Yeah that makes sense to me

AbhiPrasad: Actually we can just do pending changes length > 0

hanlinc27:
> 2 things:
> 
> 1. When i toggle the visibility of some stories, click upload, then click no on the prompt, all stories are set to is_visble=true. The weird part is published stories is fine (it still says all stories which have changed) but all stories and visible stories revert back to original form (where all stories are set to visible)
> 2. Can we also add this prompt when we click on a story then click edit story (so basically on the edit story button)

1. The first behaviour that happens I think is because of the mutation in `HANDLE_SWITCH_CHANGE` it seems like after the dialog prompt happens, the mutate doesn't stay (?!) so then it's reading the `story.is_visible` from the original `/api/stories` endpoint which is not what we want. I'm not too sure how to fix that yet though
2. Let's implement that after we can actually route to edit stories?

nimwijetunga:
> > 2 things:
> > 
> > 1. When i toggle the visibility of some stories, click upload, then click no on the prompt, all stories are set to is_visble=true. The weird part is published stories is fine (it still says all stories which have changed) but all stories and visible stories revert back to original form (where all stories are set to visible)
> > 2. Can we also add this prompt when we click on a story then click edit story (so basically on the edit story button)
> 
> 1. The first behaviour that happens I think is because of the mutation in `HANDLE_SWITCH_CHANGE` it seems like after the dialog prompt happens, the mutate doesn't stay (?!) so then it's reading the `story.is_visible` from the original `/api/stories` endpoint which is not what we want. I'm not too sure how to fix that yet though
> 2. Let's implement that after we can actually route to edit stories?

Yeah if (2) is easier to do after we route to edit stories that's fine. No rush for (1) but we should figure it out before merging this pr

DinuWije (approved):
Everything looks good to me! The only thing is I wasn't able to run it locally (not sure why, prob just be a problem on my end), so please double-check with someone else that it works on their machine before merging!

ui/src/pages/Admin/AllStories/reducer.ts
```diff
@@ -55,12 +55,13 @@ export function allStoriesReducer(state: State, action: Action): State {
       const changedVisibilityContainsID = state.changedVisibility.some(
         (e) => e.ID === action.story.ID
       );
+
+      action.story.is_visible = !action.story.is_visible;
```
Shehryar21, line 59: In Jeffery's PR, he is changing the is_visible field in the action when calling dispatch. Just coordinate with him and decide on whether to do it in reducer or do it how he did it.

ui/src/pages/Admin/AllStories/reducer.ts
```diff
@@ -78,9 +79,6 @@ export function allStoriesReducer(state: State, action: Action): State {
         return {
           ...state,
           visibleState: state.visibleState.filter((e) => e !== action.story.ID),
-          visibleTableState: state.visibleTableState.filter(
-            (e) => e.ID !== action.story.ID
-          ),
```
Shehryar21, line 83: Why did you remove the visibleTableState stuff?

Shehryar21 (changes requested):
I am getting unexpected behaviour. I am not sure if its because mutate and some visibleTableState has been removed. When I toggle off few stories and then toggle back on. Then when i switch to visible stories tab and back to all stories and then back to visible stories, those stories are duplicated in visible stories tab.

https://user-images.githubusercontent.com/37384882/115179252-e5782580-a0a0-11eb-9965-23808e2ed0a2.mov

hanlinc27:
> I am getting unexpected behaviour. I am not sure if its because mutate and some visibleTableState has been removed. When I toggle off few stories and then toggle back on. Then when i switch to visible stories tab and back to all stories and then back to visible stories, those stories are duplicated in visible stories tab.
> 
>  Screen.Recording.2021-04-18.at.11.47.02.PM.mov

Wait yeah don't review it, I already know the duplicate issue. I'm probably going to close this PR after exams and just reopen a new one with the needed changes.

ui/src/pages/Admin/AllStories/reducer.ts
liu-jeffrey: (*Jeffrey)

PR #291, opened 2021-04-10
Feat: AllStories Filtering

### LINKED ISSUE #250: All Stories Filtering - Frontend
https://www.figma.com/file/uVqnweL2ApqvPHWzQYmZ1t/Dashboard?node-id=512%3A2799

Add filtering functionality to tables.

Backend filtering here: https://github.com/uwblueprint/shoe-project/pull/237

Shehryar21:
Closes #250 . Will add more tests (just added very basic test for now) and also style the filtering button.

Please test the filtering and search and let me know if something unexpected occurs. I tested with some cases, but there could be some cases which I might have not considered.

image


restapi/stories.go
```diff
@@ -81,7 +81,7 @@ func (api api) ReturnAllStories(w http.ResponseWriter, r *http.Request) render.R
 	var storiesByTags []models.Tag
 
 	getVisibility := r.URL.Query()["visibility"]
-	visibility := true
+	visibility := false
```
hanlinc27:
```suggestion
	visibility := true
```
Since `stories` should return ALL stories now regardless of visibility

ui/package.json
```diff
@@ -23,13 +23,14 @@
   },
   "dependencies": {
     "@babel/preset-typescript": "^7.12.7",
-    "@material-ui/core": "4.11.1",
+    "@material-ui/core": "^4.0.0",
```
hanlinc27: Please revert back to `4.11.1` or `4.11.3`

ui/src/pages/AllStories/index.tsx
```diff
@@ -99,16 +145,44 @@ function createData({
...
+  const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
+
+  const open = Boolean(anchorEl);
+  const id = open ? 'simple-popover' : undefined;
```
hanlinc27: `undefined` ? Should it just be `null` or `""`?

ui/src/pages/AllStories/index.tsx
```diff
@@ -438,6 +591,18 @@ export const AllStories: React.FC = () => {
...
+            {
+              name: "tags",
+              header: "Tags",
+              width: 300,
+              onHeaderClick() {
+                handleRequestSort("tags");
+              },
+              cell: (story) =>
+                story.tags.map((tag) => (
+                  <StyledChip color="primary" key={story.id} label={tag} />
```
hanlinc27:
```suggestion
                  <StyledChip color="primary" key={story.ID} label={tag} />
```

ui/src/pages/AllStories/index.tsx
```diff
@@ -321,6 +462,18 @@ export const AllStories: React.FC = () => {
...
```
hanlinc27:
```suggestion
                  <StyledChip color="primary" key={story.ID} label={tag} />
```

ui/src/pages/AllStories/reducer.ts
```diff
@@ -1,35 +1,57 @@
-import { StoryView } from "../AllStories";
+import { StoryView } from "../AllStories/index";
 
 export interface State {
   tabValue: number;
+  search: string;
   visibleState: StoryView[];
   visibleTableState: StoryView[];
```
hanlinc27: Can you merge in `main` or rebase? There's a couple of state changes including the old search that's making the diff kind of confusing. i.e. this visibleState should be a number[]

ui/src/pages/AllStories/reducer.ts
```diff
@@ -122,5 +176,114 @@ export function allStoriesReducer(state: State, action: Action): State {
...
+      };
+    
+      const checkRowVisibility = (row) => {
+        console.log(newState.filterState.visibility)
+        return ((newState.filterState.visibility.visible && row.is_visible) || (newState.filterState.visibility.nonVisible && !row.is_visible) || (!newState.filterState.visibility.nonVisible && !newState.filterState.visibility.visible))   
+      }
+    
+      const requestSearchAndFilter = () => {
+        if (state.tabValue === 0) {
+          let filteredRows: StoryView[] = newState.origTableData.filter((row) => {
```
hanlinc27: why do you have an `origTableData` when we also have a `tableData` state?

ui/src/pages/AllStories/reducer.ts
```diff
@@ -122,5 +176,114 @@ export function allStoriesReducer(state: State, action: Action): State {
...
+            return handleTagFilterHelper(row)
+          });
+          newState.tableData = filteredRows
+        } else if (state.tabValue === 1) {
+          let filteredRows: StoryView[] = newState.visibleTableFilterState.filter(
+            (row) => {
+              return requestSearchHelper(row);
+            }
+          );
+          filteredRows = filteredRows.filter((row) => {
```
hanlinc27: Should `filteredRows = filteredRows.filter((row) => {
            return checkRowVisibility(row)
          });
          filteredRows = filteredRows.filter((row) => {
            return handleTagFilterHelper(row)
          });`

be extracted into its own helper function? A lot of repeated code here

hanlinc27 (changes requested):
Haven't gotten a chance to run it locally yet, just some initial comments.

restapi/stories.go
Shehryar21: it would not matter, but i changed it to true

ui/src/pages/AllStories/index.tsx
Shehryar21: dont think it would matter. I just copied it from https://material-ui.com/components/popover/

ui/src/pages/AllStories/reducer.ts
Shehryar21: i need it for filter / searching. origTableData has all the stories that are returned from the API call. I use that to do filtering/searching. If i use tableData, and i search or filter smth, i wont have access to previous stories.

ui/src/pages/AllStories/index.tsx
Shehryar21: I dont think this should be story.ID because then stories with multiple tags would have duplicate key and React throws an error msg for that

hanlinc27: Why can't we keep it as `key={tag}` so you know it's unique? we shouldn't be using `story.id` anywhere because that's part of the table cell id and it's confusing with` story.ID`

ui/src/__tests__/pages/AllStories.test.ts
```diff
@@ -73,15 +75,44 @@ describe("allstories table", () => {
...
 
+    it("should initialize tags and filter states after Tags API", () => {
+      const tagRows = ["TAG1", "TAG2"];
+      const action: Action = {
+        type: "INITIALIZE_AFTER_TAGS_API",
+        rows: tagRows,
+      };
+      expect(INIT_STATE.tags.length).toBe(0);
+      expect(INIT_STATE.filterState.tags).toEqual({});
+      expect(INIT_STATE.filterState.visibility).toEqual({
```
hanlinc27: hmm technically you're not expecting `filterState` to change in this test so you don't need the two expect() here.

ui/src/__tests__/pages/AllStories.test.ts
```diff
@@ -135,14 +166,28 @@ describe("allstories table", () => {
...
-        data: searchedVal,
+        type: "HANDLE_SEARCH/FILTER",
+        data: { ...INIT_STATE, search: searchedVal },
       };
       expect(INIT_STATE.search).toBe("");
       const newState = allStoriesReducer(INIT_STATE, action);
       expect(newState.search).toBe(searchedVal);
     });
 
+    it("should filter by tag1", () => {
```
hanlinc27: Since `HANDLE_SEARCH/FILTER` also changes the `visibleTableState` etc. in r`equestSearchAndFilter() `you can also test those change properly in this function

ui/src/pages/AllStories/index.tsx
```diff
@@ -133,10 +158,28 @@ function createData({
     image_url,
     video_url,
     content,
+    tags,
   };
 }
 
 export const AllStories: React.FC = () => {
+  const [anchorEl, setAnchorEl] = React.useState<HTMLButtonElement | null>(
```
hanlinc27: Put in reducer?

ui/src/pages/AllStories/index.tsx
```diff
@@ -133,10 +158,28 @@ function createData({
...
+
+  const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
+
+  const open = Boolean(anchorEl);
```
hanlinc27: nit: more descriptive name
```suggestion
  const isFilterOpen = Boolean(anchorEl);
```

ui/src/pages/AllStories/index.tsx
```diff
@@ -248,62 +327,30 @@ export const AllStories: React.FC = () => {
...
-    }
-  };
   const cancelSearch = () => {
-    dispatch({ type: "HANDLE_SEARCH", data: "" });
-    requestSearch("");
+    dispatch({ type: "HANDLE_SEARCH/FILTER", data: { ...state, search: "" } });
   };
   if (error) return <div>Error returning stories data!</div>;
   if (!allStories) return <div>Loading all stories table..</div>;
+  if (tagError) return <div>Error returning tags data!</div>;
```
hanlinc27: Do you think it's worth having a loading state for undefined on the popover for if there were tags and they didn't all fetch yet?

ui/src/pages/AllStories/reducer.ts
```diff
@@ -13,20 +13,23 @@ export interface State {
   selectedRowIds: number[];
   order: "asc" | "desc";
   orderBy: string;
+  tags: string[];
+  filterState: any;
```
hanlinc27: Let's type this out: 
```suggestion
   filterState: FilterState;
}

interface FilterState {
  visibility: VisibilityType;
  tags: Record<string, boolean>;
};
type VisibilityType = {
  visible: boolean;
  nonVisible: boolean;
};
```

ui/src/pages/AllStories/reducer.ts
```diff
@@ -161,11 +190,112 @@ export function allStoriesReducer(state: State, action: Action): State {
...
 }
+
+const requestSearchHelper = (row, newState: State) => {
+  if (newState.search == "") {
+    return true;
+  }
+  let doesExist = false;
+  Object.keys(row).forEach((prop) => {
+    //Exclude search for StoryView members not displayed on table cells
+    const excludedParameters =
```
hanlinc27: make sure to merge in changes to exclude ID from search

ui/src/pages/AllStories/reducer.ts
```diff
@@ -161,11 +190,112 @@ export function allStoriesReducer(state: State, action: Action): State {
...
+      newState.origTableData,
+      newState
+    );
+  } else if (newState.tabValue === 1) {
+    newState.visibleTableState = requestSearchAndFilterHelper(
+      newState.visibleTableFilterState,
+      newState
+    );
+  } else if (newState.tabValue === 2) {
+    console.log(newState.changedVisibility);
```
hanlinc27:
```suggestion
```

hanlinc27 (changes requested):
I really liked how you separated the functions cleanly :) I tested it with multiple tag stories and it works for me and changes persist from search and vice versa.

ui/src/pages/AllStories/index.tsx
Shehryar21: yeah kept it as key={tag}

ui/src/__tests__/pages/AllStories.test.ts
Shehryar21: I am expecting it to change, becuase once tags are fetched, I am creating a filterState which depends on the tags that were fetched by the API

ui/src/pages/AllStories/index.tsx
Shehryar21: tbh it doesnt really matter. If tags fail to be fetched, it would return an error anyways. I can keep it as `const id = "simple-popover"`

ui/src/__tests__/pages/Admin/AllStories.test.ts
```diff
@@ -73,15 +75,44 @@ describe("allstories table", () => {
       };
       expect(INIT_STATE.visibleTableState.length).toBe(0);
       expect(INIT_STATE.origTableData.length).toBe(0);
-      expect(INIT_STATE.origTableData.length).toBe(0);
+      expect(INIT_STATE.tableData.length).toBe(0);
```
AbhiPrasad:
```suggestion
      expect(INIT_STATE.tableData).toHaveLength(0);
```

AbhiPrasad: Ditto for all other `.length` assertions.

I actually usually use `eslint-plugin-jest` with https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-to-have-length.md to avoid these kinda code reviews, but too lazy to add it rn.

ui/src/pages/Admin/AllStories/reducer.ts
```diff
@@ -13,22 +23,27 @@ export interface State {
...
   | { type: "HANDLE_CHECKED_ALL"; rows: StoryView[] }
   | { type: "HANDLE_CHECKED"; e: React.ChangeEvent; story: StoryView }
   | { type: "SET_ORDERING"; order: "asc" | "desc"; orderBy: string }
   | { type: "SET_TABLE_DATA"; data: StoryView[] }
-  | { type: "SET_TAB_VALUE"; newValue: number }
+  | { type: "SET_TAB_VALUE"; newValue: number; newState: State }
   | { type: "SET_CHANGED_VISIBILITY"; data: StoryView[] }
   | { type: "SET_VISIBLE_TABLE_STATE"; data: StoryView[] }
-  | { type: "HANDLE_SEARCH"; data: string };
+  | { type: "HANDLE_SEARCH/FILTER"; data: State };
```
AbhiPrasad: As a general thing, the action shouldn't be passing in state. It should only pass in whatever parameters to change state with. As such `data: State` is a code smell for me here.

ui/src/pages/Admin/AllStories/index.tsx
```diff
@@ -122,11 +148,30 @@ function createData({
     image_url,
     video_url,
     content,
+    tags,
   };
 }
 
 export const AllStories: React.FC = () => {
+  const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
```
AbhiPrasad: let's move this function below the `useReducer()` call.

ui/src/pages/Admin/AllStories/index.tsx
```diff
@@ -122,11 +148,30 @@ function createData({
...
+      type: "SET_ANCHOR",
+      click: null,
+    });
+  };
+
+  const { data: tagOptions, error: tagError } = useSWR<string[]>("/api/tags");
+
   const [state, dispatch] = useReducer(allStoriesReducer, INIT_STATE);
+  const isFilterOpen = Boolean(state.anchorEl);
+  const id = "simple-popover";
```
AbhiPrasad: lift out of function, we don't need to declare this every component render.

```ts
const POPOVER_ID = "simple-popover";
```

ui/src/pages/Admin/AllStories/index.tsx
```diff
@@ -141,6 +186,62 @@ export const AllStories: React.FC = () => {
...
+        ...state,
+        filterState: {
+          ...state.filterState,
+          visibility: {
+            ...state.filterState.visibility,
+            [event.target.name]: event.target.checked,
+          },
+        },
+      },
+    });
```
AbhiPrasad: This logic should be done in the reducer, not the in payload given to the action.

ui/src/pages/Admin/AllStories/index.tsx
```diff
@@ -155,7 +256,10 @@ export const AllStories: React.FC = () => {
     if (allStories) {
       dispatch({ type: "INITIALIZE_AFTER_API", rows: allStories });
     }
-  }, [allStories]);
+    if (tagOptions) {
+      dispatch({ type: "INITIALIZE_AFTER_TAGS_API", rows: tagOptions });
+    }
+  }, [allStories, tagOptions]);
```
AbhiPrasad, line 316: ok to prevent this kinda stuff from happening, lets split up the data fetching from useSWR into it's own component.

```ts
export function AllStories() {
  const { data: tagOptions, error: tagError } = useSWR<string[]>("/api/tags");
  const { data: allStories, error } = useSWR<StoryView[] | undefined>(
    "/api/stories",
    fetchStories
  );

  // if tagError || error return error state;
  // if data undefined return loading state;

  return <AllStoriesInner tagOptions={tagOptions} allStories={allStories} />;
}

...

// tagOptions and allStories are always defined
function AllStoriesInner({ tagOptions, allStories }) {
  const [state, reducer] = useReducer(state, getInitState(tagOptions, allStories));
  // other hooks

  return JSX
}
```

ui/src/pages/Admin/AllStories/index.tsx
```diff
@@ -237,62 +338,30 @@ export const AllStories: React.FC = () => {
...
+      search: "",
+      filterState: {
+        visibility: {
+          visible: false,
+          nonVisible: false,
+        },
+        tags: tagBooleans,
+      },
+    };
+    dispatch({ type: "SET_TAB_VALUE", newValue, newState });
```
AbhiPrasad: again IMO the state transform should be done in the reducer, the action should only provide the variables needed to perform the transform.

ui/src/pages/Admin/AllStories/index.tsx
Shehryar21: what is the advantage of doing this over our current way of using UseEffect?

Shehryar21: also if doing this, would i create 2 files? one for AllStories(), and one for AllStoriesInner, or just separate function in the current file

AbhiPrasad: > what is the advantage of doing this over our current way of using UseEffect?

This is just a cleaner pattern, I'd rather just instantiate state with correct data rather than dispatch later on.

Read up on **functional core, imperative shell**, which IMO is a really good design pattern to leverage - https://www.destroyallsoftware.com/screencasts/catalog/functional-core-imperative-shell. Here we have an imperative shell, which relies on network IO (side effect). This is abstracted away from the functional shell -> the `AllStoriesInner` component which is just responsible for rendering some UI. This is not exactly the case, as technically `useReducer` is a side effect, but thats a discussion for another day.

Here we are setting a clear separation of concerns here. Let's say someone wanted to add another data fetching request (another GET to a endpoint), they know they just have to edit the imperative component, `AllStories`. This would involve a new `useSWR` call, and then handling the according error and loading states. They don't have to worry about mucking around in the view logic.

Think about testing as well, we can easily test either component for what it does. A more unit or UI regression test for the inner component, and a more integration based tested for the imperative wrapper component.

> also if doing this, would i create 2 files? one for AllStories(), and one for AllStoriesInner, or just separate function in the current file

Do whatever feels better to you lol

Shehryar21: i will give that a read. Another question tho. So right now in our handleSwitchChange function, we are using mutate. And after every mutate, the dispatch in our useeffect occurs i think which updates the state accordingly. Now with this new implementation, how would that change? Would mutate work fine??

AbhiPrasad: Yeah that would be fine, the props passed down would just change.

nimwijetunga (approved):
LGTM from a functional perspective, just fix the styling before you merge

ui/src/pages/Admin/AllStories/index.tsx
Shehryar21: Do you think we should do this code restructuring in this PR? Multiple people are working on this allstories stuff and might increase confusion and conflicts. It might be better to do this after functionality is complete. What do you think? I am fine doing it in this PR as well.

AbhiPrasad: Hmm, we can refactor afterwards if that works better for you.

nimwijetunga (approved):
Works fine

ui/src/pages/Admin/AllStories/index.tsx
Shehryar21: ok yeah lets do it after functionality is complete

hanlinc27 (approved):

PR #293, opened 2021-04-16
Notify users to enable cookies during login

DinuWije:
image

image

### DIFF
--- a/ui/src/pages/Login.tsx
+++ b/ui/src/pages/Login.tsx
@@ -28,7 +28,7 @@ const GoogleButton = styled.div`
   margin-bottom: 5vh;
 `;
 
-const Center = styled.div`
+const CenterIcon = styled.div`
   position: absolute;
   top: 50%;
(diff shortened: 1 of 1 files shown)


hanlinc27 (approved):
I enjoyed that last commit message :))

This is probably clear enough but maybe if we have time in the end we can do it via prompt? (https://www.npmjs.com/package/react-cookie-consent#using-it) But I guess there are only so many admin-side users.

AbhiPrasad (approved):

DinuWije:
> I enjoyed that last commit message :))
> 
> This is probably clear enough but maybe if we have time in the end we can do it via prompt? (https://www.npmjs.com/package/react-cookie-consent#using-it) But I guess there are only so many admin-side users.

Yeah I think if we have time we can put it in, but tbh cookies aren't a big deal because they are only off by default in incognito mode (it should work normally otherwise).

PR #300, opened 2021-04-24
Move Endpoints to Protected

nimwijetunga:
Since auth is now working we can move the backend endpoints to being protected

### DIFF
--- a/restapi/api.go
+++ b/restapi/api.go
@@ -70,7 +70,6 @@ func Router(db *gorm.DB, locationFinder location.LocationFinder) (http.Handler, 
 	r.Group(func(r chi.Router) {
 		r.Use(jwtauth.Verifier(config.GetJWTKey()))
 		r.Use(Authenticator)
-		// TODO: move back to protected endpoints
 		rest.PostHandler(r, "/stories", api.CreateStories)
 		rest.PostHandler(r, "/story", api.CreateStoriesFormData)
 		rest.PostHandler(r, "/authors", api.CreateAuthors)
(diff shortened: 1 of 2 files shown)


restapi/auth.go
```diff
@@ -53,16 +52,10 @@ func (api api) Login(w http.ResponseWriter, r *http.Request) render.Renderer {
...
-	if err != nil {
-		return err
-	}
-	expiration := time.Now().Add(jwtExpiry)
-
 	claim := &models.Claims{
 		Email: email,
 		StandardClaims: jwt.StandardClaims{
-			ExpiresAt: expiration.Unix(),
+			ExpiresAt: 0,
```
nimwijetunga, line 58: I set ExpiresAt to 0 so that this JWT token never expires. This is generally bad practice but the reason is that google will do auto refresh for it's own token so it may not call our backend endpoint which means that we have an outdated jwt token (if the user doesn't explicitly logout). To solve this we would need to perform auto refresh on our token as well but that is out of scope right now so doing it the hacky way for now

AbhiPrasad (approved):

restapi/api.go
```diff
@@ -57,32 +57,27 @@ func Router(db *gorm.DB, locationFinder location.LocationFinder) (http.Handler,
...
 		r.Use(Authenticator)
-
-		// rest.PostHandler(r, "/stories", api.CreateStories)
-		// rest.PostHandler(r, "/story", api.CreateStoriesFormData)
-		// rest.PostHandler(r, "/authors", api.CreateAuthors)
+		rest.PostHandler(r, "/stories", api.CreateStories)
+		rest.PostHandler(r, "/story", api.CreateStoriesFormData)
+		rest.PostHandler(r, "/authors", api.CreateAuthors)
+		rest.PostHandler(r, "/countries", api.AddCountries)
+		rest.PostHandler(r, "/countries", api.AddCountries)
```
Shehryar21: Please remove one of them. Added it twice.

Shehryar21 (approved):
Just make the one change that I commented

restapi/api.go
nimwijetunga: good catch

PR #301, opened 2021-04-26
feat: show and hide stories when checked

hanlinc27:
### Description
Give user the option to show and hide stories on batch when they select the checkmarks on the left-hand side. The multiselect frames in Figma can be found [here](https://www.figma.com/file/uVqnweL2ApqvPHWzQYmZ1t/Dashboard?node-id=1159%3A2297).
Essentially, checked boxes are now stored in two separate array states, `checkedVisibleStoriesArray` and `checkedHiddenStoriesArray`, when the user selects to either show or hide, we call `handleSwitchChange` on every single story in the respective array and dispatch the appropriate action + mutation. Please test to see if the behaviour is as you would expect. 

### Demo 

https://user-images.githubusercontent.com/19617248/116019997-6ab88880-a613-11eb-800c-473d48e04273.mov

#### To-do

- Add confirmation toast when stories are hidden or shown
- Fix the styling :apple:

### DIFF
--- a/ui/src/pages/Admin/AllStories/index.tsx
+++ b/ui/src/pages/Admin/AllStories/index.tsx
@@ -584,9 +584,6 @@ export const AllStories: React.FC = () => {
               {
                 name: "ID",
                 width: "5%",
-                onHeaderClick() {
-                  handleRequestSort("ID");
-                },
                 header: (
                   <div>
                     <Checkbox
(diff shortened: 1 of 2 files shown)


nimwijetunga (approved):
Tested this out locally and everything seems to work as expected, thanks a bunch for picking this up!

PR #335, opened 2021-04-29
Checkbox Fixes

### LINKED ISSUE #305: When Filter/Search is Active Checkbox selects all instead of just filtered
# Description

When you filter or search and it reduces the number of results when u click the checkbox to select all, it selects all stories in the table instead of just the ones that we're filtered.

image

# Steps to Reproduce

1. Filter to reduce results
2. Click the checkbox at the top row of the table

# Potential Solution

Maybe we're using the wrong state here?

### LINKED ISSUE #310: Confusing button on Visible Stories Page
# Description
Visible stories check all dropdown doesn't behave as expected (the dropdown lets you select stories that are currently not in the page). This could be an intended feature; if it is, ignore me. 

# Steps to Reproduce

A list of steps that can help someone to reproduce this bug locally

1. Go to Visible Stories page 
2. Hide one or more stories from the visible stories page
3. Using select all checkboxes on the top left of the page, select all the stories on the Visible Stories page
4. You will see a button on the top left that says Show All on Map () with the number of hidden stories in the brackets. 

I don't think this functionality makes sense for the VisibleStories page since you don't want to be able to select any hidden stories from this page.

# Potential Solution

Disable the "Show All Hidden" button from the Visible Stories page

Shehryar21:
closes #305. closes #310

Please test it thoroughly. I tried it with few simple use cases and seems to be working fine. 

One thing to note: If you checkoff few stories, and then search/filter something which results into showing no stories, and then you remove search/filter, the checked off stories go away as well. That is expected behaviour right???

https://user-images.githubusercontent.com/37384882/116511187-0e17d080-a894-11eb-9fc9-b97e05711b2e.mov

### DIFF
--- a/ui/src/__tests__/pages/Admin/AllStories.test.ts
+++ b/ui/src/__tests__/pages/Admin/AllStories.test.ts
@@ -110,23 +110,6 @@ describe("allstories table", () => {
       expect(newState.tabValue).toBe(1);
     });
 
-    it("should check one box", () => {
-      const changeEvent = {
-        target: {
-          checked: true,
-        },
-      } as React.ChangeEvent<HTMLInputElement>;
-
-      const action: Action = {
-        type: "HANDLE_CHECKED",
-        e: changeEvent,
-        story: mockTableData[1],
-      };
-      expect(INIT_STATE.selectedRowIds).toHaveLength(0);
-      const newState = allStoriesReducer(INIT_STATE, action);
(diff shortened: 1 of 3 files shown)


nimwijetunga (approved):
Tested a bunch of cases and seems to work fine for me

hanlinc27 (approved):
Tested out multiple cases and works for me :))

hanlinc27 (approved):

PR #345, opened 2021-04-30
ref(a11y): Try to address some concerns brought up by lighthouse audit

### LINKED ISSUE #170: Do an accessibility audit on the web app
Let's make sure we are doing our best a11y wise.

1) https://reactjs.org/docs/accessibility.html
2) https://www.npmjs.com/package/eslint-plugin-jsx-a11y and https://github.com/dequelabs/axe-core-npm/tree/develop/packages/react for automated testing

AbhiPrasad:
bam 
image

Fixes #170 

Fixing the performance stuff will take a bit more work but we have solid first paint so I don't think its toooo big of a deal
image

### DIFF
--- a/ui/.eslintrc.js
+++ b/ui/.eslintrc.js
@@ -10,7 +10,7 @@ module.exports = {
     "plugin:react/recommended",
     "plugin:@typescript-eslint/recommended",
     "plugin:react-hooks/recommended",
-    "plugin:jsx-a11y/recommended"
+    "plugin:jsx-a11y/recommended",
   ],
   parser: "@typescript-eslint/parser",
   parserOptions: {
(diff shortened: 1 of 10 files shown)


hanlinc27 (approved):
Obligatory run `yarn fix`? 
We're still missing a few alt tags for the styled components in `StoryDrawer` and ` PinPreview` but it seems like it's ok for accessibility.
Aside from code splitting, how could performance be (largely) improved? Is it by cutting down our bundle size with a compression tool?

AbhiPrasad:
Parcel production bundling should already minify and compress. Ways we coulda improved performance otherwise:

1) server side render and hydrate, easy if we just made this a next js service from the beginning
2) use react dev tools to debug unnecessary multiple renders - try to minimize reconciliation where possible
3) clean up console/error/info logging
4) leverage React.memo(), and useCallback and useMemo hooks
4)

nimwijetunga:
> Parcel production bundling should already minify and compress. Ways we coulda improved performance otherwise:
> 
> 1. server side render and hydrate, easy if we just made this a next js service from the beginning
> 2. use react dev tools to debug unnecessary multiple renders - try to minimize reconciliation where possible
> 3. clean up console/error/info logging
> 4. leverage React.memo(), and useCallback and useMemo hooks

These are good points if there is any low hanging fruit (esp 3) we should try address them before handoff, otherwise any changes that require refactors we should consider holding off. Don't want to add any new bugs.

PR #346, opened 2021-04-30
Fix: Map Pin Placement

### LINKED ISSUE #334: Map pins showing in ocean
image

Shehryar21:
closes #334 

image

Difficult problems require simple solutions.

### DIFF
--- a/ui/src/components/PinCluster.tsx
+++ b/ui/src/components/PinCluster.tsx
@@ -22,6 +22,8 @@ const createClusterCustomIcon = function () {
   return new L.Icon({
     iconUrl: resting,
     iconRetinaUrl: resting,
+    iconAnchor: [25, 61],
+    iconSize: new L.Point(39, 48)
   });
 };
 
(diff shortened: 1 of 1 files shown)


nimwijetunga (approved):
nice work lgtm

hanlinc27 (approved):

hanlinc27:
With this change, the pin cluster pins are the same size as the individual pins which we don't want visually though?

nimwijetunga:
I think it's better than having pins in the water though @jenna-hirji thoughts?

hanlinc27:
@nimwijetunga @Shehryar21 try changing the resting states for pin in `Pin.tsx` to something slightly smaller than L.Point(39,48) if that's your cluster size.

Shehryar21:
I can change the cluster size to whatever we want. What is the size we want?

hanlinc27:
> I can change the cluster size to whatever we want. What is the size we want?

Let's just use the Figma hover state: 
width: 48px;
height: 57.93px;

PR #349, opened 2021-05-01
fix: table row height

### LINKED ISSUE #328: All story information should wrap
# Description
The story information should all be displayed in the table. If it's too long it shouldn't be truncated with [...] it should wrap. This is especially important when it comes to tags since it can be confusing for users when they filter by a certain tag but it isn't shown to them. 

In the example below, I filtered by Sheryar tag but you can't see it in the tags in the table.
Screen Shot 2021-04-27 at 10 14 48 PM

# Steps to Reproduce
1. Log in
2. Look at the table
3. If you want to recreate the tags example specifically, filter by the tag "Sheryar"

# How to fix
IDEK LOL - design

# Notes

This is how they should wrap:
Screen Shot 2021-04-27 at 10 18 09 PM

ALSO!!! Add more padding. It's hard to read when the columns run into each other.

hanlinc27:
Increased the table row height for all rows globally in order to show two lines of tags.
image
Closes #328

### DIFF
--- a/ui/src/components/VirtualizedTable.tsx
+++ b/ui/src/components/VirtualizedTable.tsx
@@ -21,6 +21,12 @@ const useVirtualizedTableStyles = makeStyles({
       fontSize: fontSize.subtitle,
       lineHeight: "120%",
       marginRight: "none",
+      height: "auto",
+      "& div": {
+        whiteSpace: "normal",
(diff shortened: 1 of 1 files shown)


nimwijetunga (approved):
lgtm

PR #350, opened 2021-05-01
Bug fix: filter on map page only shows countries from visible stories

### LINKED ISSUE #341: Filter showing countries not visible on map
The filter on the main map currently shows countries that are attached stories that are set to be not visible.
image

DinuWije:
## Description
- Current behaviour: The filter on the homepage page currently shows countries that are attached to stories that are not visible on the map (ie. you can filter by the country, but when you do nothing shows up on the map)
- Created a fix + ran a short test

### DIFF
--- a/restapi/authors.go
+++ b/restapi/authors.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/biter777/countries"
 	"github.com/go-chi/render"
-	"github.com/thoas/go-funk"
 	"github.com/uwblueprint/shoe-project/internal/database/models"
 	"github.com/uwblueprint/shoe-project/restapi/rest"
 )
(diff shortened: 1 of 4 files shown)


restapi/authors.go
```diff
@@ -37,24 +36,14 @@ func (api api) CreateAuthors(w http.ResponseWriter, r *http.Request) render.Rend
...
+	var countries []string
 
-	for _, entry := range authors {
-		if entry.OriginCountry != "" {
-			countries[entry.OriginCountry] = true
-		} else {
-			api.logger.Debug("Empty origin country")
-		}
-	}
+    err := api.database.Table("stories").Where("is_visible = true").Distinct().Pluck("author_country", &countries).Error
```
nimwijetunga, line 41: Can you leave a comment here indicating the SQL query:
`SELECT DISTINCT author_country from stories where is_visble=true`

Gorm is not that intuitive so this could be helpful for people reading this code

nimwijetunga (approved):
lgtm

restapi/authors.go
AbhiPrasad, line 41: Does the `Distinct()` call make sure there are no duplicate countries here?

restapi/endpoints_test.go
```diff
@@ -345,6 +345,24 @@ func (suite *endpointTestSuite) TestGetAllStories() {
 	response.Object().Value("payload").Array().Element(1).Object().Value("current_city").Equal("Toronto")
 	response.Object().Value("payload").Array().Element(1).Object().Value("tags").Array().Empty()
 }
+
+func (suite *endpointTestSuite) TestGetVisibleCountries() {
```
AbhiPrasad, line 349: I'd like to see a test to make sure that no duplicate countries are returned.

restapi/authors.go
nimwijetunga: Yeah the equivalent query is `SELECT DISTINCT author_country from stories where is_visble=true` (https://gorm.io/docs/advanced_query.html#Pluck) but i agree the test should check for this

restapi/endpoints_test.go
DinuWije: I think this test covers it because there are two stories with the Country as France, but only one of them shows up in the returned array. Note that array size is confirmed to be 3, not 4 (showing that there are no such duplicates) 
--> `[Canada, France, US]`

PR #353, opened 2021-05-01
Try/Catch Youtube URL Parsing

nimwijetunga:
Preview will work in the case that the youtube url is broken. If the user tries to upload the backend will fail and the user will get a message


AbhiPrasad (approved):
Yeah I mean it would be nice to write a test with this but whatever lets 🚢 🇮🇹

PR #337, opened 2021-04-29
feat: warning pages on close tab and styling changes

hanlinc27:
### Description
Add an event listener to the `window` to check when user reloads, closes tab, or wants to navigate back. This prompts this warning (default to google chrome): 
Screen Shot 2021-04-29 at 3 03 38 PM
Closes #314 
Changed the styling of the chip and change image buttons on upload/edit stories.
image
image
Closes #315

PR #358, opened 2021-05-01
fix: Load before redirect

AbhiPrasad:
Running the redirect before the leads to some awkward transition states, this aims to fix that.

### DIFF
--- a/ui/src/pages/Admin/AllStories/index.tsx
+++ b/ui/src/pages/Admin/AllStories/index.tsx
@@ -1,6 +1,5 @@
 import { Grid } from "@material-ui/core";
 import AppBar from "@material-ui/core/AppBar";
-import { CircularProgress } from "@material-ui/core";
 import Button from "@material-ui/core/Button";
 import Checkbox from "@material-ui/core/Checkbox";
 import Chip from "@material-ui/core/Chip";
--- a/ui/src/pages/Admin/AllStories/index.tsx
+++ b/ui/src/pages/Admin/AllStories/index.tsx
@@ -1,5 +1,6 @@
 import { Grid } from "@material-ui/core";
 import AppBar from "@material-ui/core/AppBar";
+import { CircularProgress } from "@material-ui/core";
 import Button from "@material-ui/core/Button";
 import Checkbox from "@material-ui/core/Checkbox";
 import Chip from "@material-ui/core/Chip";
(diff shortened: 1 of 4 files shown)


nimwijetunga:
hmm it still seems as though when i make changes (don't publish) then logout those changes still persist. Honestly this probably not too much of a big deal since this edge case should be quite rare so if the fix is not simple then might not be necessary

AbhiPrasad:
> hmm it still seems as though when i make changes (don't publish) then logout those changes still persist. Honestly this probably not too much of a big deal since this edge case should be quite rare so if the fix is not simple then might not be necessary

~Oh yeah sorry I was getting to that. It happens because we actually do a `mutate` (PUT) request everytime we flip a switch, see: https://github.com/uwblueprint/shoe-project/blob/main/ui/src/pages/Admin/AllStories/index.tsx#L411-L428.~

~Should we remove the eager mutate? I don't think we want that to happen right?~

Nvm that was cap, idk whats going on lol

hanlinc27:
I believe it's because of https://github.com/uwblueprint/shoe-project/blob/3937505e502bf37e19340a445c9c543d5f067e69/ui/src/pages/Admin/AllStories/index.tsx#L376
https://swr.vercel.app/docs/options
which was needed to prevent the page from making a request even when we pressed "cancel" for the prompt asking about navigating away. For instance, commenting that out means that when we log out and log in the changes don't persist.

nimwijetunga:
> I believe it's because of https://github.com/uwblueprint/shoe-project/blob/3937505e502bf37e19340a445c9c543d5f067e69/ui/src/pages/Admin/AllStories/index.tsx#L376
> 
> https://swr.vercel.app/docs/options
> 
> which was needed to prevent the page from making a request even when we pressed "cancel" for the prompt asking about navigating away. For instance, commenting that out means that when we log out and log in the changes don't persist.

Did it work when you commented it out @hanlinc27?

hanlinc27:
> > I believe it's because of https://github.com/uwblueprint/shoe-project/blob/3937505e502bf37e19340a445c9c543d5f067e69/ui/src/pages/Admin/AllStories/index.tsx#L376
> > 
> > https://swr.vercel.app/docs/options
> > which was needed to prevent the page from making a request even when we pressed "cancel" for the prompt asking about navigating away. For instance, commenting that out means that when we log out and log in the changes don't persist.
> 
> Did it work when you commented it out @hanlinc27?

Yes but then the prompt cancel still sends a request when the warning shows up.https://github.com/uwblueprint/shoe-project/pull/337

PR #361, opened 2021-05-03
fix iconAnchor for individual pin

Shehryar21:
Currently when zoomed out and individual pin is showing (not a pin cluster), the pin does not stay in its original position. For example in the pic the story city is Toronto, but it did not stay at Toronto when zoomed out.

Before:
image

Now:
image


hanlinc27 (approved):

nimwijetunga (approved):
lgtm!

